PR #19, opened 2024-02-24
Adds internationalization

courtneypattison:
The translated files are not completed yet and I need to add a redirect from /calibrate because that route is broken now in favour of [locale]/calibrate.


hluedeke (commented):
It's looking good! I did not spot any code issues - the only potential issue might be if there is text somewhere that got missed, but I'm sure we will hear about it if there is anything that gets missed!

hluedeke:
From what I can tell, things look good, but it’s very hard to tell since files were renamed and those whole files are showing as new. This one probably makes more sense testing from a browser. If things look good on your Vercel build, I’d say go for it!

I can pull the branch and run it locally later today or tonight, but if you don’t want to wait for me, that’s fine too!

courtneypattison:
It does touch too many files! I'll be happy to have this one merged. Here's the [vercel preview](https://pattern-projector-git-intl-courtney-pattisons-projects.vercel.app?_vercel_share=myPomsJeoGCSUYBUlj1zKB8fdnFzdJqD) so you don't have to build it locally.

hluedeke:
We appear to have lost the border around the opened pdf. Everything else appears to be intact!

courtneypattison:
I'm not sure how I managed to put that change in there! I moved it to the inner div so it rotates with the pdf.

PR #20, opened 2024-02-24
Resolves #17. Adds logic based on the location of the cursor to displ…

### LINKED ISSUE #17: Hover cursor should change within the radius of a hotspot
If the cursor is within the radius of a corner, the cursor should be different than the cursor in drag mode. This will make it obvious where the hotspots are, even if they are off.

hluedeke:
…ay the target symbol when close to the corner and a pan symbol anywhere else on the calibration grid. Should make it easier to detect hotspots


hluedeke:
How can I see why a Vercel deployment failed? Do I need to have permission? If I click on the details, it gives me a 404.

courtneypattison:
Sorry, I’m not sure how to give permissions for seeing build errors. I’ll look into that.  It only runs yarn build so you can at least test that locally!

hluedeke:
@courtneypattison that's good enough for me! You don't need to bother with Vercel permissions

courtneypattison:
Okie doke! When do you think I should be merging into main/deploying (since they are linked right now)? Should I do it as soon as a branch is ready, like this one? Or should I try to batch them?

hluedeke:
🤷‍♀️ I'm not sure, to be honest. What makes the most sense to you?

hluedeke:
I have a branch for feature-15, but I have not opened a PR yet because I actually branched 15 from 17 (easier to test with hover cursor). I know you asked earlier about how to merge things. Would you rather I PR things against the new look branch? Or an hluedeke branch? This one's up to you, however you'd like to organize it!

courtneypattison:
Hmm, I think I'll merge into main whenever it looks like nobody is online. Also, I think a hluedeke branch makes sense. The garbage files are from next-pwa. I add them to [the .gitignore in the intl branch](https://github.com/Pattern-Projector/pattern-projector/pull/19/commits/f37caee9179c334b881d688b853aa7e5223576b5), but that's not merged yet!

PR #25, opened 2024-02-26
Add Dutch translation

roel-v:
Dutch translation of all strings in en.json .


courtneypattison:
Thank you so much @roel-v!

Issue #26, opened 2024-02-26
Add contribution documentation


courtneypattison:
Add steps for running the project locally.

PR #31, opened 2024-02-26
Add developer quick start guide (for issue #26)

roel-v:
I started writing some developer notes for the Electron stuff I was working on and included some getting started there as well. This version has the Electron things stripped out again so as to be relevant to main. It's high level enough to assume that a user can figure out how to install git and what the path in the environment is. I can flesh out further if you want.


courtneypattison:
This is awesome, thank you! I think the amount of detail is great. Just one little thing: the project uses yarn as a package manager, so maybe add something like:

- Install yarn

  $ npm install --global yarn

- Install all packages required for Pattern Projector:

  $ yarn

- Start development webserver:

  $ yarn dev

instead of the npm install and npm run dev.

Thank you again!

roel-v:
OK I'll look into it.

On a related note, didn't want to open an issue but there is a package-lock.json that got checked in a few days ago; is that correct? Isn't package-lock.json managed by npm? If I do 'yarn install' I get a warning about this file being there and about not mixing package managers.

hluedeke:
@roel-v if you commit to your own main, then it will reflect in a PR from your main to the project's main automatically. It's actually rather convenient that it works this way!

If you are doing translations that you don't want to mix with your other technical commits, I'd highly recommend maintaining two separate branches, such as ```i18n``` and ```main```.

You get a new branch by doing the following via command line:

```git checkout -b i18n```

Keeping a separate branch up to date requires a little more upkeep, but it's not bad.

Before you start coding, ideally you make sure all your branches are up to date first. You would do this by the following steps:

1. From Github, make sure your fork is synced with the main project (you can do this from the Code tab - it will let you know you are out of sync).

2. From your command-line:

```git checkout main```
```git pull origin``` <= syncs your local code with your forked project in GH
```git checkout i18n```
```git merge main``` <== this makes sure i18n has all the updates from your main branch and should now be up to date

You could probably add this to your developer documentation (which could arguably be yet another branch, haha). Let me know if you have questions, and I'll do my best to answer!

hluedeke:
> Also, I woke up this morning thinking we should add a line about running $ yarn build before making a PR because PRs are automatically deployed as a vercel preview which is run with a yarn build.

Yes, this has become part of my PR process!

roel-v:
Thanks @hluedeke - I've now realized I need to be much more cautious about being on the right branch, I've shot myself in the foot a few times already. It seems that in order to rescue this PR I will have to do a rebase which is beyond my current understanding. So I'm just closing this one and making a new one.

About the "yarn build" - is that just to check that it still works as it should and that the vercel task doesn't get tripped up? Or does it do more?

hluedeke:
> Thanks @hluedeke - I've now realized I need to be much more cautious about being on the right branch

Sorry if I overexplained that - I figure too much information is better than not enough. Hope it was helpful!

> About the "yarn build" - is that just to check that it still works as it should and that the vercel task doesn't get tripped up? Or does it do more?

Yes, that's exactly it.

roel-v:
> Sorry if I overexplained that - I figure too much information is better than not enough. Hope it was helpful!

Oh it's been most helpful, there are so many ways to do things, it's hard to see the git forest for the trees sometimes.

roel-v:
Continued in https://github.com/Pattern-Projector/pattern-projector/pull/39 .

PR #35, opened 2024-02-26
React pdf custom renderer - NOTE: this will probably break in Vercel

hluedeke:
I don't think I got all the Typescript definitions right on this, so I don't think Vercel will build. But you should be able to pull it locally and run a dev server from it.

Best of luck!

PS. This does not include the code that accesses line thickness, only layers

### DIFF
--- a/app/_components/pdf-custom-renderer.tsx
+++ b/app/_components/pdf-custom-renderer.tsx
@@ -1,0 +1,86 @@
+import { useEffect, useMemo, useRef } from "react";
+import invariant from "tiny-invariant";
+import { pdfjs, useDocumentContext, usePageContext } from "react-pdf";
+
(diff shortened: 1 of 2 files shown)


hluedeke:
I think my plan was going to be to create a new context provider in the pdf-viewer component with logic for toggling layers - maybe pass a function with the layers as an argument? Then you could theoretically use useContext() in the custom render to implement the logic. I'm not sure if that will work, but it was my initial thought.

You'll also notice that the custom renderer doesn't scale correctly, so that will also need to be fixed, unfortunately. When I realized it didn't match the original PageCanvas class in the library, I got a little discouraged with the approach.

courtneypattison:
I got the scaling working :) I'll take that as a win for now and look into this more later today.

courtneypattison:
Closing in favour of layers branch

PR #40, opened 2024-02-27
Add layer toggling

courtneypattison:


hluedeke:
I feel that description 😂

hluedeke:
Is Typescript broken just because you are returning null on that optionalContentConfigPromise? I think that might be documented incorrectly in the underlying library, because it's typed as "required" but the comments say to return null in some cases...

I noticed that the other day.

courtneypattison:
Lol, I noticed that as well! That error message was bizarre. It pretty much said don't return the `config | null` but if you return `null` you'll get a default config. So goofy.

TypeScript is worried that page is not defined. It seems like tiny-invariant checks for false which seems to be was react-pdf returns if there's no page proxy but it doesn't let typescript know it's not undefined.

courtneypattison:
TODO: look at this example of usePageContext hook with TypeScript next time I’m on the laptop. https://github.com/wojtekmaj/react-pdf/blob/93b09c3b0a4a231d401c5aa3f8f375f746d5c0d2/packages/react-pdf/src/StructTree.tsx#L15

hluedeke:
I believe this is the actual canvas renderer we are trying to mimic:

https://github.com/wojtekmaj/react-pdf/blob/93b09c3b0a4a231d401c5aa3f8f375f746d5c0d2/packages/react-pdf/src/Page/PageCanvas.tsx

courtneypattison:
that's a good reference. thank you!

hluedeke:
Wow, I never thought to try putting CustomRenderer in a callback - does that really work? That's amazing if it does!

courtneypattison:
Lol, I think it's working. I'm not confident though. If you have a chance to check out the vercel preview, please do!

courtneypattison:
@roel-v If you have the time, could you pretty please translate a few new items for layer toggling?
```
"layersOn": "Show layer menu",
"layersOff": "Hide layer menu",
"recenter": "Recenter PDF"
```
I'd love it if you could look over the [preview](https://pattern-projector-git-layers-courtney-pattisons-projects.vercel.app?_vercel_share=oHdivXZxtSCYInQLCbcmkN7rXWyQ5VYn) too to see if you can find any issues!

roel-v:
Hope it's OK I just put the translations here as it's just a few - I don't have much time today:

"layersOn": "Toon lagen menu",
"layersOff": "Verberg lagen menu",
"recenter": "Centreer PDF opnieuw"

courtneypattison:
Perfect, thank you!

courtneypattison:
Also, I totally forgot to ask for a translation for 
```
"fourCornersOn": "Corner Highlights On",
"fourCornersOff": "Corner Highlights Off",
```
When you get the time, sorry! I need to get more organized!

roel-v:
"fourCornersOn": "Alle hoeken markeren",
"fourCornersOff": "Alleen geselecteerde hoek markeren",

This says, translated back into English "Highlight all corners" and "Only highlight selected corner", which IMO is a better description of this functionality; I did it this way because there is no idiomatic 1:1 translation.

courtneypattison:
Agreed! I’ll change that in en. Thank you!

PR #45, opened 2024-02-29
Adds more key management. Moves arrow key handler for calibration can…

hluedeke:
…vas into react hook that uses progressive movement. Changes the colors of the canvas corners to follow rainbow (a precursor for rotation support). Updates tab handler to use tabIndex, eg. allows all buttons on the page to receive tab focus before tabbing to grid. Once in focus, the tab button will infinitely select corners. Pressing the escape key will defocus the calibration grid and reset tabindex back to the top of the page. Removed default pointToModify corner, as tabIndex should go through all the buttons first before selecting a corner (@courtneypattison do you agree with this approach?)

Also allows for modification of the "sensitivity" of progressive arrow translation. The reason for this is the calibration grid corner translation needs to be less sensitive than PDF translation.

I'd like to get a second pair of eyes on the update to the PDF translation logic. I'm pretty sure "localTransform" does not get saved anywhere, but when I refactored the arrow key handler into a hook, I modified the offset translation to only be present in ```setMatrix3d()```, as opposed to modifying the ```localTransform``` object directly. It does not appear to affect usage on the page, but this works differently than it used to. I just want to be sure this is an acceptable change for the long run.

### DIFF
--- a/app/[locale]/calibrate/page.tsx
+++ b/app/[locale]/calibrate/page.tsx
@@ -22,6 +22,7 @@ import {
 } from "@/_lib/transform-settings";
 import { CM, IN } from "@/_lib/unit";
 import useProgArrowKeyHandler from "@/_hooks/useProgArrowKeyHandler";
+import useProgArrowKeyToMatrix from "@/_hooks/useProgArrowKeyToMatrix";
 
 const defaultPoints = [
   // Points that fit on an iPhone SE
@@ -178,9 +179,11 @@ export default function Page() {
     }
   }, []);
 
+  const pdfTranslation = useProgArrowKeyToMatrix(!isCalibrating);
+
   useEffect(() => {
-    setMatrix3d(toMatrix3d(localTransform));
-  }, [localTransform]);
+    setMatrix3d(toMatrix3d(localTransform.mmul(pdfTranslation)));
+  }, [localTransform, pdfTranslation]);
 
   useEffect(() => {
     if (points && points.length === maxPoints) {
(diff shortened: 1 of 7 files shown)


courtneypattison:
I'll look at this tomorrow! I'm up way past my bedtime lol.

courtneypattison:
> …vas into react hook that uses progressive movement. Changes the colors of the canvas corners to follow rainbow (a precursor for rotation support). Updates tab handler to use tabIndex, eg. allows all buttons on the page to receive tab focus before tabbing to grid. Once in focus, the tab button will infinitely select corners. Pressing the escape key will defocus the calibration grid and reset tabindex back to the top of the page. Removed default pointToModify corner, as tabIndex should go through all the buttons first before selecting a corner (@courtneypattison do you agree with this approach?)
> 
Yes! I think that makes sense. It's much better integrated with the rest of the page now.

> Also allows for modification of the "sensitivity" of progressive arrow translation. The reason for this is the calibration grid corner translation needs to be less sensitive than PDF translation.
> 
> I'd like to get a second pair of eyes on the update to the PDF translation logic. I'm pretty sure "localTransform" does not get saved anywhere, but when I refactored the arrow key handler into a hook, I modified the offset translation to only be present in `setMatrix3d()`, as opposed to modifying the `localTransform` object directly. It does not appear to affect usage on the page, but this works differently than it used to. I just want to be sure this is an acceptable change for the long run.

Looks good to me! localTransform is such a bad name! I think adding pdfTransform makes it more clear what's going on.

PR #47, opened 2024-03-01
Converts CustomRenderer back to a Class in pdf-viewer. Passes layers/…

hluedeke:
…setLayers down through context instead. This fixes the flickering for me.

### DIFF
--- a/app/_components/pdf-custom-renderer.tsx
+++ b/app/_components/pdf-custom-renderer.tsx
@@ -2,12 +2,14 @@ import {
   Dispatch,
   SetStateAction,
   useCallback,
+  useContext,
   useEffect,
   useMemo,
   useRef,
 } from "react";
 import invariant from "tiny-invariant";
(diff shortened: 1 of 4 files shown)


courtneypattison:
I've never seen this done before! I need to work on my React knowledge! I just memoized the callback like you suggested and I think that works as well. I don't know enough about React to decide which one makes more sense if they both work.

hluedeke:
They probably both work. I can test the memoized version if that’s easier to read for you and we can commit that version. This is using react context, which is basically a fancy way to pass props from an ancestor component to any of its descendants. It’s really cool when you need to pass the same props to large parts of the application (for example, you can pass a variable from the url or a themeing object all the way down your component tree without passing it as an arg to each component).

You’ll notice react-pdf uses it as well with getPageContext().

courtneypattison:
Ah, that's cool. Good to know!

courtneypattison:
Closing since useCallback seems to fix this issue and it's already in the layers branch.

Issue #49, opened 2024-03-01
PDF Page switching: give visual cue that user is on the first/last page


roel-v:
The back/fwd buttons on the page switching tool don't change color to indicate when 'back' is unavailable (when on the first page) or 'fwd' is unavailable (when on the last page). The mouse cursor does turn into a cursor to indicate the widget is no longer clickable. To make things consistent with how disabled widgets tend to work, the icon needs to switch to a 'dimmed' state, e.g. to a lighter grey color.

courtneypattison:
That's a good idea!

hluedeke:
I'm gonna knock this one out in the tooltips branch, since I will be refactoring the header component anyway.

courtneypattison:
I’m so excited to see it refactored! Thank you!

PR #56, opened 2024-03-02
Resolves #21 and #49. Adds rudimentary tooltips. Refactors header buttons to IconButtons. Updates show layer menu logic.

### LINKED ISSUE #21: Add tooltips for icon buttons
Maybe on hover show a tooltip showing what the icon buttons do. I’ve gotten a few messages where people aren’t sure what they are for e.g. “ I'm trying to delete a projection and start over, but I can only move the current grid. Is there a way to delete and start over.”

hluedeke:


courtneypattison:
This looks so much better! Thank you!

A few small things:
- The full screen button hasn't been updated to an IconButton. This probably shouldn't be a component anyway. I'm not sure what I was thinking on this one. Do you think it should be pulled out of its own component and just be added to the header?
- The info page icon doesn't have a tooltip
- What do you think about making the github icon on the landing page an IconButton with a tooltip? Something like "see source code" or "see pattern projector code." I'm having trouble thinking of something short and informative.

hluedeke:
1. I saw the FullScreen button - it does have different logic/styling than the other buttons, so I just left it alone. If you'd rather do the IconButton, I'm good with that. If so, you can update or I can
2. Info page icon - yeah, I wasn't sure what to put there. I also didn't want to bother with translations, haha. So I left that one be. We can add whatever you want there
3. To be honest, I haven't even touched the landing page. I didn't even look to see if there were any buttons in there that could be icon buttons. If you think it fits the bill, then sure, let's convert it. You could put a tooltip there, but I think anyone who knows what Github is isn't going to need a tooltip, haha.

PR #71, opened 2024-03-07
Menu hide/show

hluedeke:
I made some changes to the original PR:

1) Default the menu to hidden in full-screen mode (this feels intuitive and I think even expected)
2) Show/hide is only available in full-screen mode
3) Show/hide button now has a border
4) Show/hide button is now in header tag, making the transition more natural
5) Removes containing tooltip, which creates a hidden non-draggable area on the screen

Let me know what you think. Based on the new video in the PFS page, it appears it is expected for the menu to disappear in full-screen mode and to be available again in non full-screen mode. Chrome tells you to hit escape in full-screen mode to get out of it, so I'm hoping most people will understand that that is how full screen works. The button is there in case they do not.

### DIFF
--- a/app/_components/header.tsx
+++ b/app/_components/header.tsx
@@ -131,13 +131,21 @@ export default function Header({
     }
   }, [isCalibrating, setShowLayerMenu]);
 
+  useEffect(() => {
+    if (fullScreenHandle.active) {
+      setShowNav(false);
+    } else {
+      setShowNav(true);
+    }
+  }, [fullScreenHandle.active]);
+
   return (
     <>
       <header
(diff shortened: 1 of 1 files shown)


courtneypattison:
Love it! So much better! Only one tiny thing: the icon is slightly off centre because the left-1/2 is relative to the left side of the button. Do you want write access to this project? It's starting to feel a bit silly copying PR branches back and forth when we could just be working on the same branch. If you don't though, I totally get it!

hluedeke:
Sure 👍 I can merge mine into your ```menu-hide``` branch and we can work off of that?

courtneypattison:
Okie doke, I think you have write access now. Just an FYI, merging into main also deploys to patternprojector.com. I should probably setup more steps than that but for now, that's how deploying to production works.

hluedeke:
Ok, I probably won't merge into main then unless directed to :)

PR #79, opened 2024-03-09
Fixes #69. Automatically turns grid off when uploading a pdf file

### LINKED ISSUE #69: Make it more obvious that you can cut outside of the grid on the project page
People aren't sure if they can cut outside of the grid on the project page. Now that we have major and minor lines for measuring, should we remove the border on the grid and make it go to the edges of the screen?

hluedeke:


courtneypattison:
Thank you!

PR #83, opened 2024-03-10
Larger Text, Attempt 1

hluedeke:
This applies the text layer at correct scale. It's rudimentary at best, so I wanted to build it for you to take a look at. The competition for screen space is going to be the biggest issue, I think. Even with hover, sometimes it's hard to find the exact text you are looking for if there is a bunch of text in the same small space.

I'd like feedback here on a direction, as it's better than nothing, but not great.

Thanks!

### DIFF
--- a/app/_components/calibration-canvas.tsx
+++ b/app/_components/calibration-canvas.tsx
@@ -149,7 +149,6 @@ function drawLine(
   lineWidth: number = 1,
 ): void {
   ctx.beginPath();
-  console.log(lineWidth);
   ctx.lineWidth = lineWidth;
   ctx.moveTo(p1.x, p1.y);
   ctx.lineTo(p2.x, p2.y);
--- a/app/_components/pdf-viewer.tsx
+++ b/app/_components/pdf-viewer.tsx
@@ -70,7 +70,7 @@ export default function PdfViewer({
       itemIndex: number;
       transform: Array<any>;
     }) => {
-      return `<span class="opacity-0 hover:opacity-100" style="background-color: #FFF; color: #000; font-size: 60px; letter-spacing: 5px">${str}</span>`;
+      return `<span class="opacity-0 hover:opacity-100 hover:text-6xl" style="background-color: #FFF; color: #000; font-size: 60px; letter-spacing: 5px">${str}</span>`;
     },
     [],
   );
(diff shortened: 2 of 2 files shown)


courtneypattison:
I feel like this is so much better than what we have right now! It's not perfect but at least the text is legible now.

I have an idea, but haven't tried it at all. What about overlaying a scaled up version of the PDF viewer that is clipped to a magnifying glass size around the mouse that would have a transform origin set to where the mouse is? If you think that's crazy, no need to look into it! I'd be happy adding this PR as is. Or I could look into this idea when I wrap up a few other things.

hluedeke:
I'm completely tracking with you there. I was already thinking about copying the Document/Page elements from pdf-viewer and seeing if we could work with a clipped version of the pdf - perhaps this would be it's own custom renderer, like zoom-box-renderer or something. I agree that this is probably the way to go in the long run!

It's funny you were thinking the same way. I started to comment about it a few hours ago but was hesitant because I'm not sure how much work it's going to be.

courtneypattison:
Cool beans, I’m glad we’re on the same PDF page, lol. I’d be nice if it works!

hluedeke:
Agree. I’m ok getting this version out as a stopgap until we have time to properly flesh out the zoom box. The one thing I’d maybe do is abstract out the PDF_TO_CSS_UNITS to a helper class, but idk when I’ll have a chance to do it.

I created a new branch for this in case you wanted to merge there and do your own edits. Otherwise, it’s not a huge amount of code (so not a lot to refactor), so I’d be ok deploying as is too.

courtneypattison:
Okie doke, as soon as a I get a chance, I think I’m going to merge and clean up all the current PRs and just get them out there! It’s trivial to rollback deployments, so if I’ve broken something, I can easily go back. I’ve got a vet visit and a kiddo home on March break though, so things are slow going. March break started on Friday and he’s been wanting to use PP everyday. He recalibrates every time for fun lol. We’ve made a stormtrooper quilt block and the tailors ham that Roel linked to a few days ago. Good find @roel-v

hluedeke:
>  so things are slow going

I beg to differ 😅 I feel like a slacker!

> he’s been wanting to use PP everyday. He recalibrates every time for fun lol

This is wonderful ❤️

PR #85, opened 2024-03-10
Global checkbox for layers menu

hluedeke:
Also squared off the menu corners (I would like to "attach" this menu to the top menu at some point). Move the show/hide layers button to the left side of the screen instead of the top menu because in full screen mode, it's harder to access layers with the main menu hidden.


courtneypattison:
This makes so much more sense! If you have a chance, please feel free to mess with the stitch menu placement. I moved it to the left but it still feels goofy.

hluedeke:
Ok, sounds good! I'm sorta better, but not really, so I'm going back to sleep for awhile. It's a bummer to be sick over a weekend, haha.

courtneypattison:
That sucks big time. Sleep well!

Issue #68, opened 2024-03-06
On touch devices, provide on-screen arrow buttons to move selected handle


roel-v:
I don't know if it's possible to detect if the user has a keyboard or not; if not, maybe this should be a toggleable option. But when a corner handle is selected, show four arrow keys on the side of the display (maybe elsewhere? trade offs) that act as a limited onscreen keyboard. I find per-pixel movement much easier and faster than dragging but I can't get a keyboard to show up when doing this on my Android tablet. Having an on screen built-in alternative is I think the solution here.

courtneypattison:
Agreed, the tablet experience is not as nice right now. Maybe a next button in the middle to go to the next corner. I'm loving tabbing through the corners and adjusting with the arrow keys.

Also, any ideas for showing tooltips on the touch screen? I'm not sure how other software does this. onClick with a timeout?

hluedeke:
As far as I know, tooltip support isn't common in mobile - usually mobile has a different layout, which typically uses screen size  breakpoints to determine. I think we will need more than that here (will probably use user-agent) because smaller screen on a desktop will definitely not want visible buttons. I can take a stab at this one.

j-huff:
Something that could be interesting would be a fine-adjust mode when dragging the selecting points on touch screen. Like tapping and holding the selecting circle without dragging for a few moments will activate it, and then your dragged movements will only move the selection circle at like a 5:1 ratio or something. Could help with calibration on touch screens.

courtneypattison:
> Something that could be interesting would be a fine-adjust mode when dragging the calibration points on touch screen. Like tapping and holding the circle without dragging for a few moments will activate it, and then your dragged movements will only move the calibration point at like a 5:1 ratio or something. Could help with calibration on touch screens.

I really like this idea! The main issue I could see with it is if someone doesn't realize they've activated this mode and get super frustrated with movements being slowed down. Maybe a button or something to toggle it?

PR #89, opened 2024-03-11
Added precision mouse/touch movement. 

j-huff:
If the user presses down on a calibration point and doesn't move more than 4 canvas units in 500ms, precision movement mode is activated. The calibration point will appear as a crosshair instead of circle, and the movement of the point will be at a 5:1 ratio to the mouse movements.

Values for delay, movement threshold, and movement ratio are adjustable. The values I picked seem right to me, but you can play around with them before merging if you'd like. Cheers.

### DIFF
--- a/app/_components/calibration-canvas.tsx
+++ b/app/_components/calibration-canvas.tsx
@@ -420,6 +420,8 @@ export default function CalibrationCanvas({
       setPoints(localPoints.map((p) => applyOffset(p, dragOffset)));
       setDragOffset({ x: 0, y: 0 });
       setPanStart(null);
+    } else {
+        setPoints(localPoints);
     }
     setIsPrecisionMovement(false);
     setDragStartTime(null);
@@ -436,6 +438,7 @@ export default function CalibrationCanvas({
 
   function handleTouchUp() {
     localStorage.setItem("points", JSON.stringify(localPoints));
+    setPoints(localPoints);
     setPointToModify(null);
     setPanStart(null);
     setIsPrecisionMovement(false);
--- a/app/_components/calibration-canvas.tsx
+++ b/app/_components/calibration-canvas.tsx
@@ -20,7 +20,7 @@ import { CornerColorHex } from "@/_components/theme/colors";
 import useProgArrowKeyPoints from "@/_hooks/useProgArrowKeyPoints";
 
 const maxPoints = 4; // One point per vertex in rectangle
-const PRECISION_MOVEMENT_THRESHOLD = 4;
+const PRECISION_MOVEMENT_THRESHOLD = 15;
 const PRECISION_MOVEMENT_RATIO = 5;
 const PRECISION_MOVEMENT_DELAY = 500;
 
@@ -160,7 +160,6 @@ function drawLine(
   lineWidth: number = 1,
 ): void {
   ctx.beginPath();
-  console.log(lineWidth);
   ctx.lineWidth = lineWidth;
   ctx.moveTo(p1.x, p1.y);
   ctx.lineTo(p2.x, p2.y);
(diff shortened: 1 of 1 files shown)


j-huff:
Unclear why the deployment failed on vercel. Vercel just give me a 404 when I try to access the details. Works fine in my development environment 🤷

hluedeke:
Try running ```yarn build``` in your local environment. That should let you know what the issue is.

courtneypattison:
Thank you! This solution feels a lot more elegant than adding in arrow buttons to move corners around. I'm not completely sold on how it's activated though. I'm having trouble intentionally activating it. Is there a way we could make it more obvious? Also, here's the [preview](https://pattern-projector-git-fork-j-d2c88c-courtney-pattisons-projects.vercel.app?_vercel_share=6yV0mAEMSOxo5PoD0oT0pzh38t1kOiAB) for the other folks.

j-huff:
> I'm having trouble intentionally activating it. Is there a way we could make it more obvious?

Hmm. Are you using a touchscreen? So far I've only tested on my 1080p screen with a mouse. We may have to increase PRECISION_MOVEMENT_THRESHOLD for it to be easy enough to activate for touch screens. It's set to 4 canvas units right now, meaning you have to keep your cursor within 4 canvas units of where you clicked for 500ms for precision movement to activate.

Is there any reason why the screen to canvas conversion would differ significantly for different devices? It may be the case that the distance calculation should be done using screen units instead if that's the case.

j-huff:
It might be an issue with your cache. I just noticed that it would stop working sometimes if I changed something in the code, and would only start working again once I cleared my cache in chrome. Even doing a control+shift+r reload wasn't working. Had to fully clear my chrome cache and site data. Not sure if this is an actual issue with the code, or just weird behavior when reloading a different version of the same component.

courtneypattison:
Thank you! I just pulled this one. I'm going to try to use it as a base for cleaning up the calibration canvas.

courtneypattison:
I pulled it and I think it works great! @hluedeke Did you work on #68? I saw you self-assigned that one. Do you think this would be a good replacement for that or maybe in addition to?

hluedeke:
I started it, but it was going to need a refactor of the key handler hooks I'd already written for easing. At the time, I was indecisive on exactly how I wanted to refactor it, so I didn't get terribly far and moved on to other things. If this fixes the issue, let's use this!

courtneypattison:
Okie doke, we'll start with this and if it still feels like the tablet experience is worse than the desktop/laptop experience we can still add the buttons in.

courtneypattison:
I went to merge this one into main today and tested it on the tablet. Unfortunately the calibration canvas isn't showing. I'm guessing it's a performance issue because it's an older iPad and that may happen if there are too many state changes invalidating components. I'll try to root cause this more later!

j-huff:
Is that possible performance issue happening only on this branch?

courtneypattison:
I haven't checked on your other branch, sorry! I think if you refactored this one keeping in mind that setState is asynchronous, avoiding using state as much as possible, and maybe memoizing with useCallback when you need to this would work beautifully.

j-huff:
> I haven't checked on your other branch, sorry! 

No, I meant is this performance issue happening on the main branch as well? Not that I would be surprised that my nooby React skills would cause performance degradation 😅 I'll have to revisit this with performance in mind.

courtneypattison:
Ah, sorry, I misunderstood! The main branch seems good on the iPad :) React is funky. I feel like I'm constantly goofing up using it.

j-huff:
Using the React Developer tools I didn't notice anything about this branch in particular being a performance issue. calibration-canvas.tsx barely takes any time to render and I don't think I changed anything that would effect its parent component (page.tsx). However, there is plenty of performance benefit to be gained. Because the "points" and "perspective" states are declared in page.tsx and modified every mouse movement in calibration-canvas.tsx, the entire page is having to be rerendered every frame the mouse is moving a calibration point. This is an issue as we shouldn't be re-rendering slow-to-render components such as the header so often. If we want performance gain, we should decouple "points" and "perspective" (and maybe some other states) in calibration-canvas and only update the points/perspective states of page.tsx when we actually need to. The easiest way to do this would be to only update points/perspective at the end of the drag sequence.

j-huff:
yeesh, I guess I didn't test that last commit very thoroughly >.<

courtneypattison:
Thank you @j-huff! I'll test this one on the iPad tomorrow!

courtneypattison:
I used this PR as the base for the beta branch. After reseting the calibration grid, it looks great. I thought I tested that before but I must have misremembered, sorry!

PR #90, opened 2024-03-11
Adds stitch menu

courtneypattison:
I rebased from main instead of merging and found myself in a sticky situation. Here's the new version of the stitch menu. Do we think this is good enough for prod? Am I dreaming?

Here's the [preview](https://pattern-projector-git-stitch-menu-courtney-pattisons-projects.vercel.app?_vercel_share=FfHPy6ggSWhdlC1W8hiiBhAjcTchZdio)!


hluedeke:
I'm finally getting around to testing this out - I think it's great for a first go! We might want to tweak it a bit in the future, but it's a great start, and so cool!

I think you could release now if you wanted, or we can tweak it a bit first, if you rather.

hluedeke:
When testing, I noticed that the grid columns didn't save. After doing a quick peek through of your code, I didn't see any saving happening. Are you wanting to table that completely for now?

courtneypattison:
I'm really uncertain about the saving. Do you think we should should at least save by file? Have an object in localStorage like `stitchSettings` with filenames as keys and all the values from the stitch menu? Maybe add a count to start writing over old ones after a max is reached.

hluedeke:
One thing you might consider is file names that don’t qualify as keys. I’m not sure what the restrictions are for Map, but objects (and likely json) will have a few. We can either do a regex-like replace on them (and have a slight risk of file name duplicates), or we can do an array of objects with two keys: fileName and stitchSettings (or whatever).

I wonder if the File object provides a compressed hash or other unique identifier for each file based on contents🤔 that could be another way that isn’t dependent on file name.

courtneypattison:
I still haven't added this *facepalm* I'm feeling so whishy washy on it. I think regex like replace with risk of filename duplicates makes sense. I'm hoping people won't really expect it to be saved. I'm just not sure about saving things and then having to deal with backwards compatibility when the inevitable new fields get added to the stitch menu.

hluedeke:
Ok, I don't want to push you anyway you don't want to go - I'm not sure I follow on the "new fields". Are you referring to saving presets or something else?

courtneypattison:
Lol, you're not pushing at all! I just mean that PDFStitcher has a ton of options for different cases and I'm sure the stitch menu won't work for a lot of those cases so I'm guessing we'll need to add in more options, maybe row or column order? I'm not really sure.

courtneypattison:
That's a good idea!

hluedeke:
Glad I sold you on it 😁 it's not uncommon to do stuff like that, as long as you apply sensible defaults. And with Typescript, it will whine at you if you are missing something!

PR #97, opened 2024-03-14
Axis-locked Constrained drag + mouse out during drag bug fix

j-huff:
When holding the control key, the PDF pattern will only be moved exactly along the X or Y axis (whichever part of the vector is bigger). The key can be pressed before or after the drag has begun.

Also added a fix for when the mouse moves off the element when dragging. Before, the movement would stop being updated because it was attached to the onMouseMove event of the element rather than the window. The onMouseMove event of the window gets used now instead which fixes this issue. There may be a way to achieve a similar react while still using React.MouseEvent, but this was the simplest solution IMO.

### DIFF
--- a/app/[locale]/calibrate/page.tsx
+++ b/app/[locale]/calibrate/page.tsx
@@ -362,7 +362,8 @@ export default function Page() {
           setTransformSettings={setTransformSettings}
         />
         <Draggable
-          className={`cursor-grabbing select-none ${visible(!isCalibrating)}`}
+          viewportClassName={`select-none ${visible(!isCalibrating)}`}
+          className={`select-none ${visible(!isCalibrating)}`}
           localTransform={localTransform}
           setLocalTransform={setLocalTransform}
           perspective={perspective}
--- a/app/_components/draggable.tsx
+++ b/app/_components/draggable.tsx
@@ -91,11 +91,11 @@ export default function Draggable({
       const dest = transformPoint(p, perspective);
       const tx = dest.x - dragStart.x;
       const ty = dest.y - dragStart.y;
-      var vec = {x: tx, y:ty};
+      let vec = {x: tx, y:ty};
       if (isAxisLocked){
         vec = toSingleAxisVector(vec);
       }
-      let m = translate(vec);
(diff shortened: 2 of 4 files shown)


courtneypattison:
Thank you for working on this! Today I learned that control + click on a Mac is the same as right click on Windows, lol. So this is bringing up the context menu for me. What do you thinking about changing it to 'Shift'?

You can check out the calibration canvas to see how to use React's [onKeyDown](https://github.com/Pattern-Projector/pattern-projector/blob/3f12521ff5a11db9389842c5c27b5c97f4741a30/app/_components/calibration-canvas.tsx#L294). Also, I think maybe wrapping draggable in div with full viewport height and width and setting the onMouseEvent there would be better than adding listeners to the window.

j-huff:
> So this is bringing up the context menu for me. What do you thinking about changing it to 'Shift'?

Can do

> Also, I think maybe wrapping draggable in div with full viewport height and width and setting the onMouseEvent there would be better than adding listeners to the window.

Yes I had a similar thought. I'll try working that in.

j-huff:
Also changed the keybind to shift instead of control. I think this implementation is a little cleaner. The only downside to using this sort of viewport approach is that you can no longer drag your cursor outside the window and have the drag position still update. This is probably a pretty rare use case though as most people would be using fullscreen, and probably wouldn't need to be dragging something so far in a single drag.

There's a weird issue with react-pdf overwriting the cursor style when dragging that I couldn't figure out a good solution to.

courtneypattison:
Cool beans! Thank you! I've had to patch react-pdf before to update scaling. You can see it in the patches directory. We don't need that patch anymore because we ended up making a custom renderer, but if you see where the cursor is set in react-pdf, you might be able to update the patch to set it to the cursor that you'd like instead. Just an idea though! I know its not fun wading through a library trying to patch their stuff.

j-huff:
Thanks, didn't know patching like that was a thing. Drag cursor correctly now. BTW I had to remove the .next directory to get the style to update.

courtneypattison:
You're fast!

PR #102, opened 2024-03-15
Applies user unit to pdf scale, should fix large pdf scale issue

hluedeke:


hluedeke:
I seem to have introduced an issue that causes the menus to clip under the large PDF when it is dragged over them. This does not seem to be happening with a regular sized PDF.

hluedeke:
Ok, well - I decided to try and square up my calibration canvas to see if I could figure out at exactly what point the menus stop rendering...and now they are rendering fine. It seems to trigger on heavy calibration grid skew. Maybe it's due to the matrix3d css transform?

courtneypattison:
Looks great! I'm not seeing the the menu clipping on my computer (Chrome/Mac). It has made me notice another issue though #104. It's such a weird number of duplicates, lol.

hluedeke:
I have a Chrome/Mac too. I strongly suspect it has to do with the matrix3d transform in css, since lightly skewed calibrations are not an issue. I'm guessing we're seeing it now because the scale is so large...but I'm happy to call it an edge case and see if anyone else complains :)

I would look at #104, but I don't have a PDF with grouped layers to test with.

courtneypattison:
I think you're probably right that the size is making things a bit wonky. These PDFs are silly big. A lot of designers have made projector files with huge margins for Adobe and now it's causing trouble for us.

courtneypattison:
The pattern that I'm using doesn't have grouped layers either: this is the one I'm using to test https://petitestitchery.com/collections/freebies/products/adult-jack-zip-front-hoodie I haven't tried any of their other patterns to see if it's a common occurrence.

PR #105, opened 2024-03-15
Resolves #104. Squishes layers into a Map keyed by name instead of id

### LINKED ISSUE #104: Duplicate layers in Petite Stitchery patterns
I was just testing out PR #102 and it's working great for scaling but it surfaced another issue. The layer menu has around 5 duplicates of the same layers. It could have to do with #50. I never got around to checking for layer grouping but when I open the same pattern in Adobe I don't see any grouping. These Petite Stitchery patterns are troublemakers!

hluedeke:
This does not remove layers that do not have content, but it removes duplicate layers by condensing files with multiples of the same layer names to a single named layer with multiple ids.

### DIFF
--- a/app/_components/pdf-custom-renderer.tsx
+++ b/app/_components/pdf-custom-renderer.tsx
@@ -14,6 +14,7 @@ import type {
   PDFDocumentProxy,
 } from "pdfjs-dist/types/src/display/api.js";
 import { Layer } from "@/_lib/layer";
+import { PDFPageProxy } from "pdfjs-dist";
 
 function erodeImageData(imageData: ImageData, output: ImageData) {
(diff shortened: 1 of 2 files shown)


hluedeke:
We have it implemented how PDFStitcher does today (verified that this morning), so I'd say let's just get it out there. I'm not even 100% sure how we're going to detect whether a layer has objects in it yet anyway.

hluedeke:
> Ah, why are PDFs so messy?! 

This pretty much applies to any protocol or standard that's been implemented ever, lol - definitely not unique to PDFs! If it's allowed, people will do it!

courtneypattison:
You're so right! Merging now.

PR #106, opened 2024-03-15
Add dimension labels to calibration canvas

### LINKED ISSUE #92: Grid orientation issue
I don’t have a fix for this one off the top of my head, but a user in PFS flipped their calibration height/width measurements (aka entered them backwards). This caused confusion as their projected file was very distorted.

I think the confusion was due to the same issue(s) we have seen when people have a rotated projector image from what they see on their screen.

courtneypattison:
Fix #92 :) This one keeps coming up


hluedeke:
Looks great! I was not picturing labels inside the grid, but I think it's a good idea since many people use the whole space to project.

Issue #110, opened 2024-03-18
Small visual bug when switching units


j-huff:
When switching between cm and in, there's a single frame in which the canvas is redrawn incorrectly. This is low priority as I don't image people would be switching units like this, but it's a little jarring and should be fixed regardless.

j-huff:
Seems to be happening because the canvas dedraws a frame in which ptDensity has been updated, but the localPerspective hasn't been yet. Fix will involve making sure localPerspective and ptDensity always agree with eachother in the draw function.

hluedeke:
Are both ptDensity and localPerspective listed as dependencies on the draw function? It's possible that that function has too many dependencies causing unnecessary re-renders. I've seen that in at least one other place and corrected it. We may just need to trim the dependency list (on useEffect/useCallback/useMemo, whichever applies).

courtneypattison:
You are so fast! I'll check out the fix now :) Thank you!

PR #112, opened 2024-03-18
Visual bug fix for when units are changed.

j-huff:
Fix for issue #110 

Perspective matrix is now calculated in the same useEffect as the when the draw function is called. Removed localPerspective state as it is unnecessary. This should also improve performance during drag, as each time a point was moved, the canvas was being drawn multiple times: once for when localPoints changed, and again when that tickled down to localPerspective updating. This fix makes sure the canvas is only redrawn when necessary.

Simply updating the dependency list for the draw useEffect wasn't adequate and introduced other visual bugs.


courtneypattison:
So much cleaner, thank you!

PR #113, opened 2024-03-18
Beta

courtneypattison:
New Dutch translations. Thank you @roel-v! Trying to figure out a sensible Weblate workflow.


j-huff:
Should we make our pull requests to the beta branch, or the main branch? When creating feature branches, should we branch off the beta branch?

courtneypattison:
I think beta would be best. I can always switch the pr base to main if there’s other stuff on beta that we don’t want to deploy yet. I’m trying to batch changes a bit more so I’m not breaking the production build as often. I’m open to different ideas though!

Issue #114, opened 2024-03-18
Mouse movement during page refresh causes calibration points to reset


j-huff:
There's a bug where if you refresh the page while you are moving the mouse, the calibration points can reset to their default position. This is likely due to the decoupling of the points in the CalibrationCanvas, and some weirdness that happens during page load. Possible fix would be to prevent localPoints from updating points until the points have been loaded from localStorage? I'll look into it.

courtneypattison:
I noticed this before picking up my son from school so I looked into it a little bit. I think maybe checking if pointToModify is null and returning early in the touch and mouse up handlers might fix it. I didn’t get a chance to test though and won’t be able to go on my computer until later when the kiddos are asleep.

PR #115, opened 2024-03-18
Concave polygon visual bug fix + other visual bug fixes

j-huff:
Changes:

- When the calibration points create a convex polygon, the grid and text are no longer rendered. Instead, a checkboard fill pattern is rendered to indicate that the points are invalid. This prevents some of the visual wonkiness when the perspective transform is invalid.
- Created app/_lib/drawing.ts to hold canvas drawing functions to keep calibration-canvas.tsx from becoming excessively long.
- Bug fix for issue #114 : handleMouseUp short circuits if there's no drag to finish.
- Bug fix for single frame page load visual bug: remove default points (replacing with an empty array). This ensures the calibration page is not rendered until it gets "proper" points. Everything seems already set up to allow this change.

### DIFF
--- a/app/[locale]/calibrate/page.tsx
+++ b/app/[locale]/calibrate/page.tsx
@@ -29,14 +29,6 @@ import { EdgeInsets } from "@/_lib/edge-insets";
 import StitchMenu from "@/_components/stitch-menu";
 import FlexWrapIcon from "@/_icons/flex-wrap-icon";
 
-const defaultPoints = [
-  // Points that fit on an iPhone SE
-  { x: 100, y: 300 },
-  { x: 300, y: 300 },
-  { x: 300, y: 600 },
-  { x: 100, y: 600 },
-];
-
 export default function Page() {
   // Default dimensions should be available on most cutting mats and large enough to get an accurate calibration
   const defaultWidthDimensionValue = "24";
@@ -45,7 +37,7 @@ export default function Page() {
 
   const handle = useFullScreenHandle();
 
-  const [points, setPoints] = useState<Point[]>(defaultPoints);
+  const [points, setPoints] = useState<Point[]>([]);
   const [transformSettings, setTransformSettings] = useState<TransformSettings>(
     getDefaultTransforms(),
   );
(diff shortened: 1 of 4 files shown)


courtneypattison:
This looks awesome! I love the new checkerboard. Calibration canvas is so much cleaner and I’m happy to be rid of the iPhone SE points, lol. That was some hackiness from prototype time.

I’m only looking at this on my phone, so I may be missing it. Is there an early return on the handleTouchUp? That could set the points in localStorage to be an empty array now right? If the loading points from localStorage or default is slow.

j-huff:
Ah, yeah I forgot to handle that as well. Easy fix. However, I am getting some weird endless loading bug when using my chrome browser in touch emulation mode. Like the react component hits and infinite loop somewhere? Could have to do with the checkerboard pattern reference possibly. will investigate.

courtneypattison:
Lol, it’s always fix one thing break another with development eh?

j-huff:
Seems so lol. I seem to have broken panning as well 🤔

j-huff:
Disregard the whole "infinite load". That seems to be my chrome browser just being weird when attempting to do touch emulation. So unless you can recreate it, I don't think it's an actual issue. This branch should be okay now.

j-huff:
Was panning on touch screens ever working?

j-huff:
A decent amount of visual bugs have been caused by race conditions in the drawing useEffect, due to multiple items in the dependency array needing to update at the same time.

Going forward, we should consider combining inherently linked state items such as points and perspective. Doing this would prevent this type of race condition.

courtneypattison:
> Going forward, we should consider combining inherently linked state items such as points and perspective. Doing this would prevent this type of race condition.

Yes, I need to be more careful about this!

> 
> Edit: Another way around this type of issue is to use references. Like you would create a reference to perspective and use perspectiveRef.current in the useEffect, leaving only points in the dependency array. Something like this might make more sense for states which are only semi-linked. Like they change together, but not always.

I wouldn't have thought of this idea and I like it!

>I think in general, just reducing the number of state variables is the easiest thing. Like how I removed "perspective" from calibration-canvas, and just calculate it from points when needed.

Yes, let's focus on this one. Fewer state variables makes for fewer state race conditions.

roel-v:
This too is such a nice polish feature. Nice.

Issue #62, opened 2024-03-04
Add dark mode


courtneypattison:
Requested by someone who uses PP. I originally hard dark/light modes but removed the dark mode because I found calibrating easier with the light around the grid being white from the projector. I think that it might make sense to go back to having light/dark mode and make the calibration grid lines green since that seems be the easiest way to see. With tailwind it's pretty easy to add in, I just want to make sure it doesn't make calibration harder for people in dark mode. Thoughts?

hluedeke:
Hard to say. I saw the preview with the green lines, and though I haven't tried it on the projector, I think I prefer the black/white vs. the black/green as @roel-v mentioned in a comment somewhere. Has only one person requested this? It feels like one of those issues where we wait and see if others have similar requests.

courtneypattison:
Yeah, only one request for dark mode and another one for different colour calibration grid. We could always have the invert colours button on the calibration page as well so they could toggle the calibration colours depending on the colour of their mat. Or even if we keep just black and white for the grid and have dark mode support that honours their browser preferences and/or add a toggle for dark/light mode.

PR #116, opened 2024-03-19
Add dark mode based on invert colors

courtneypattison:
Finally getting around to this one!

Add #62. Only inverts to white. Maybe have a green option? Not sure if thats too funky lol. Maybe at least for the calibration grid?

Also, it doesn't affect the landing page. Not sure if I should update that as well since the toggle isn't there. Maybe set landing page colour scheme based on system settings? Is that too confusing?


hluedeke:
Yes! I can see the preview!

I think it's less important to do the main page...people click to that once then spend most of their time on the calibration screen. I wouldn't bother with the main page.

courtneypattison:
Ah, I'm glad the previews are finally working!

Issue #70, opened 2024-03-06
Add a simple one line measuring tool


courtneypattison:
I had to make a few adjustments and cut a few simple rectangles today. It would have been nice to have a simple measuring tool that measures from one point to another for figuring out the sizes of the pieces. For rectangles, I prefer just using my ruler rather than the pattern piece for a straighter edge.

Inspired by [Adobe PDF Reader](https://helpx.adobe.com/ca/acrobat/using/grids-guides-measurements-pdfs.html)

hluedeke:
Are you thinking of drawing your own rectangles with a predetermined size? Or are you saying you want to click on one location in the canvas then click on another location then have the distance be calculated? If the second, would you want snapping?

courtneypattison:
The latter, snapping optional. I usually just want a rough estimate. What do you think?

hluedeke:
Sure 🤷‍♀️ works for me. Maybe a ruler icon button that toggles the mode on/off. When mode is on, show crosshair cursor, one click starts the line, second click does the measure. Third click removes the previous measurement and starts a new one (?). Right-click cancels a measurement (?). IconButton or even hotkey (m key maybe?) would toggle the mode.

courtneypattison:
That’s exactly what I was thinking!

courtneypattison:
It could show the current measurement while moving the cursor after the first click.

hluedeke:
That works too! I think we already have the orthogonal logic built in with that geometry library. So this feels pretty straightforward to implement.

courtneypattison:
Yeah, I think so too. There are always little gotchas with the pdf versus css pixels though!

roel-v:
As I was testing two days ago, I used Foxit pdf reader a bunch of times to double check measurements too, having this build into PP would be great (what I tested was measuring on the projected image with my measuring tape, then checking against measuring that same distance in the pdf viewer; I was off by maximum 2 mm, and 90% of the time 1mm or less. I feel that's more accurate than tracing on paper, cutting and then tracing again on the fabric, but proving that feeling would require some more structured testing). Anyway, one thing that I found/find useful in Foxit is the ability to constrain the angle at which measuring is happening by holding down the shift key. This restricts measuring to horizontal, vertical or at a 45 degree angle. When moving, a line is shown between the location the user clicked first and where the cursor is currently, so that you see what distance you're measuring. Holding shift basically snaps that line to the nearest of those 3.

courtneypattison:
[...]

PR #117, opened 2024-03-19
Add measuring

courtneypattison:
Adds some of #70 I still need to add key handlers and constraining to angles.

### DIFF
--- a/app/[locale]/calibrate/page.tsx
+++ b/app/[locale]/calibrate/page.tsx
@@ -27,6 +27,7 @@ import Tooltip from "@/_components/tooltip/tooltip";
 import { useTranslations } from "next-intl";
 import { EdgeInsets } from "@/_lib/edge-insets";
 import StitchMenu from "@/_components/stitch-menu";
+import MeasureCanvas from "@/_components/measure-canvas";
 
 export default function Page() {
   // Default dimensions should be available on most cutting mats and large enough to get an accurate calibration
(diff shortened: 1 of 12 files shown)


courtneypattison:
I think I've added all the suggested functionality except hotkey toggling. I'm not sure where we want to put that. @hluedeke Do you have any suggestions for where we should put the all the hotkeys for the menubar? I believe you suggested before putting it in one place.

hluedeke:
I think we should have a keydownlistener (or keyuplistener, pick your poison) and use a react hook to define it. That way we minimize the number of places that keydown handlers are called. The implementation would be very similar to the arrow key progressive handlers, where the handler is called in the hook, but the callback is accessible from anywhere that uses the hook.

The benefits here are a single keydown handler in the project on the window, and *hopefully down the line, simplification of troubleshooting collisions on key press, as key management is handled from a single place. The hook returns a callback, and the callback can be used anywhere in the project that we need key management.

courtneypattison:
I tried to add what you described. Please let know if I've understood you correctly! Also, thoughts on this PR? Is it ready for beta?

hluedeke:
> I tried to add what you described. Please let know if I've understood you correctly! Also, thoughts on this PR? Is it ready for beta?

Yes! We'll probably want to roll in the existing arrow key handlers to use the same callback at some point, but at least everything is in the same area of code at this point. Calling document-level handlers from nested components can be a nightmare when the component hierarchy gets complicated enough. I think this looks great! (I haven't run it yet in preview) I look forward to trying it!

courtneypattison:
Awesome :) I'm quite happy with the useKeyDown hook. I think that was a great idea. We're slowly cleaning things up!

hluedeke:
> Awesome :) I'm quite happy with the useKeyDown hook

Yeah! Looks great! This might seem like an unnecessary abstraction, but when you are troubleshooting things, the ability to lookup hook usages and clearly/cleanly read them will be very helpful I think. And I think we plan to introduce a lot more hotkeys, so getting this organized up front is a good idea!

hluedeke:
Looks good!

courtneypattison:
Now just merge conflicts lol!

PR #118, opened 2024-03-19
Dark-mode canvas transition

j-huff:
With some tinkering I was able to implement animated canvas fill and stroke color transitions. Tried to match it as best as possible to how the header transitions. The projection page background also does an animated transition now as well (viewport div coming in handy).

The only things left that do not do a smooth color transition is the button background colors, and the PDF itself (may not be worth the effort? I haven't touched the PDF renderer stuff).

### DIFF
--- a/app/[locale]/calibrate/page.tsx
+++ b/app/[locale]/calibrate/page.tsx
@@ -353,7 +353,7 @@ export default function Page() {
             setTransformSettings={setTransformSettings}
           />
           <Draggable
-            viewportClassName={`select-none ${visible(!isCalibrating)}`}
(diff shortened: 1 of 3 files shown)


courtneypattison:
Ooo that’s smooth! I’m not sure it’s worth it to do the buttons or pdf. It looks way better as is. Up to you! I’m happy to merge this as is. Thank you!

j-huff:
> Ooo that’s smooth! I’m not sure it’s worth it to do the buttons or pdf. It looks way better as is. Up to you! I’m happy to merge this as is. Thank you!

Yeah, after messing with it for a bit, I've decided it might be best to just leave it alone for now. I won't be making any more commits to this branch, so feel free to merge away 🫡

roel-v:
This is really cool and makes things feel so much more polished.

Issue #67, opened 2024-03-06
Allow for explicit language selection


roel-v:
I posted this as a comment in another issue but I think that got lost in the ruckus.

Right now, the language that is used is (from what I can tell from react-intl docs) derived from the Accept-Language header. The user is then redirected from / to /[locale-id]. This page (well it's the same page but routed differently) then decides on locale by looking at the URL.

However, as most people who are not native English speakers on the web will attest, it is quite common that either the Accept-Language header is wrong for whatever temporary or permanent reason, or that you just want to see the page in another language. For example it's quite common that translations are so bad that it's more comfortable to just use English (in the case of PP and Dutch that's something that's on me...), or sometimes it's the other way around and you'd rather see the page in your local language even if you use English language versions of OS and browser and hence have your default Accept-Language set to en. Another reason is when you're using guides or tutorials in one language but you use it in another language, and you need to switch to match or check vocabulary.

Personally I can easily change the URL to switch but it would be very helpful to non-native English speakers if there is a prominent way to switch languages manually. The most obvious being a dropdown list with supported languages in the top banner somewhere. This should also work in the app itself ('/calibrate'), and preferably work without page reload. (for the 'let me switch quickly while I'm following along this Youtube vid' use case).

courtneypattison:
Great idea!

PR #120, opened 2024-03-20
Partial implementation of #67, let user switch language manually

roel-v:
This implements the desired behavior of https://github.com/Pattern-Projector/pattern-projector/issues/67 , but I didn't know how to the styling; basically, the components needs to be right-aligned (I think) and use the same visual style as the rest of the site. Can anyone point me to how to do those things?


roel-v:
Screwed up what commits are included in this.

PR #121, opened 2024-03-20
Partial implementation of #67, let user switch language manually

roel-v:
This implements the desired behavior of https://github.com/Pattern-Projector/pattern-projector/issues/67 , but I didn't know how to the styling; basically, the components needs to be right-aligned (I think) and use the same visual style as the rest of the site. Can anyone point me to how to do those things?

### DIFF
--- a/app/[locale]/page.tsx
+++ b/app/[locale]/page.tsx
@@ -27,7 +27,7 @@ export default function Home() {
       <nav className="flex items-center gap-2">
         <PatternProjectorIcon ariaLabel="" />
         <span className="font-bold">{t("beta")}</span>
-        <LanguageSwitcher ariaLabel={t("language")} />
+        <LanguageSwitcher ariaLabel={t("choose-language")} />
         <Tooltip description={t("github")} className="ml-auto">
           <IconButton href="https://github.com/Pattern-Projector/pattern-projector">
             <GithubIcon ariaLabel={t("github")} />
--- a/app/[locale]/page.tsx
+++ b/app/[locale]/page.tsx
@@ -27,8 +27,8 @@ export default function Home() {
       <nav className="flex items-center gap-2">
         <PatternProjectorIcon ariaLabel="" />
         <span className="font-bold">{t("beta")}</span>
-        <LanguageSwitcher ariaLabel={t("choose-language")} />
-        <Tooltip description={t("github")} className="ml-auto">
+        <LanguageSwitcher ariaLabel={t("choose-language")} className="ml-auto"/>
+        <Tooltip description={t("github")}>
           <IconButton href="https://github.com/Pattern-Projector/pattern-projector">
             <GithubIcon ariaLabel={t("github")} />
           </IconButton>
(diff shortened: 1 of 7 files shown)


courtneypattison:
Thank you! This looks great :)

The styling is done with tailwind, so you have to use classNames to set the style on the element e.g. `className="ml-auto"` This sets the left margin to auto which is what you'll need to add to the language switcher to right align it. Check out the classNames of other components you'd like to borrow styles from and copy them over. I think it looks great as is though, lol.

It may make sense to put the languages in the messages directory with the other translations, like they do in this [next-intl example](https://next-intl-example-app-router.vercel.app/en). What do you think? 

To get the vercel preview to deploy, you have to make sure there aren't any build errors. Running `yarn build` will give you any errors. In this case, you just need to remove the ariaLabel for the switcher!

Sorry it took so long to get to this! It's been a hectic day for me.

roel-v:
Sorry I knew about the yarn build thing at some point but didn't remember. Anyway from what I can tell, that's all fixed now, and I also fixed the styling. I reused the InlineSelector component, and I did remove the explicit width w-20 style from it which (from what I can tell) was there to explicitly size the units dropdown. I think it all looks good without it, you just may want to double check.

roel-v:
Oh another thing about putting the languages in translatable strings. I've done it but I don't think these should be translated, I think you always want those strings in the native language that it targets, e.g. 'Deutsch' for German and not 'German'. I don't know if there is a way to annotate these strings so that this is shown to translators in weblate? I put them as translatable strings because it felt conceptually cleaner, but when I think about it again, I guess it's conceptually much cleaner still to not make strings that shouldn't be translated translatable.

It looks like Weblate supports such hinting for some file formats but not JSON key/value pairs, I can't find in Weblate itself either if there is a way to add notes to strings there, like stored externally from the actual JSON files. Let me know how you want to handle this.

courtneypattison:
Looks great!

I agree the inline select for units looks good without the width. It even makes more sense now because the height was always shorter, now the width is too.

I didn't think about language in the current language versus the actual language, oops.

Screenshot 2024-03-21 at 2 41 32 PM

What about both, like this? If not, I think it may make more sense to just keep it out of the translation files so people don't translate it by accident. What do you think?

roel-v:
i've had this discussion a number of times and every time there was someone involved with a strong opinion on it, so I always just let them have it because I don't feel it one way or the other lol. On the one hand, I can see theoretically why you'd show both. But if I look at how often I have gained anything as a user from having that extra information, it's 0. The only thing is that as a user I need to be able to find back a language I know if I accidentally change it to, say, Russian, because I can't read Cyrillic and then I'm stuck with trying them all if everything switches to Cyrillic all of a sudden. That's accomplished with just having the strings hard coded in their native language. So I'm fine with anything but if do you want me to decide, I'd revert this part and go back to hard coded single strings.

As an aside (because I write more asides in issue trackers than I write actual code), this is an ever-recurring issue that some people have strong feelings about and where it will be wrong for some people no matter what you choose. Some will argue we shouldn't even have used just a language identifier, but we need proper locales, because en-US is not like en-GB is it now and OMG there isn't an en-UK WTFBBQ. and I'm sure there will be people who take offense that I chose a US flag to represent 'English' (and a Dutch flag for Dutch, but luckily I have the ethnic background where I can decide on that one without being culturally assuming or inappropriate :) ). I've seen the most inane complaints about this topic many years ago in another job, but hey if even FAANG have to spend millions managing this issue, I don't feel guilty making *some* choice and then throwing my hands up.

courtneypattison:
I agree completely! Let's go back! Sorry for the extra work.

roel-v:
No worries, easy fix.

courtneypattison:
Awesome, thank you!

PR #125, opened 2024-03-21
vcenter the height of the text on the canvas

roel-v:
This one irrationally bugged me - the height of the canvas was drawn at the middle of the left edge, without taking the actual text height into account, so the bottom of the text was at the middle, not the middle. I've tested this on current Firefox and Chrome on Windows, I don't know about support on older devices for measureText and actualBoundingBoxAscent. Is there anyone with such a device who can test?


hluedeke:
Nice! I don't have an older device, but it looks great on what I do have!

courtneypattison:
They both seem well supported according to MDN, so I'm happy to add this one in. Thank you for keeping PP classy!

Issue #124, opened 2024-03-21
Set rotate and flip origin based on click/touch


courtneypattison:
I've been thinking about this one for a while but I'm not sure how to make it simple but obvious what it going on. My current thoughts are maybe the cursor on the project page could be a crosshair that slices the screen from top to bottom and left to right so it's obvious that it is doing something special. Then the last place that the user clicks/touches is the origin for the following rotation or flip. That way people can flip on a fold line rather than flipping around the centre of the PDF. I'm not happy with this solution but thought I'd put it out there as an idea.

hluedeke:
My initial thought here is, why not just make the current mouse position the origin and introduce hotkeys for flipping/rotation?

This is an advanced feature anyways, and I feel like advanced users would like pick up what’s going on. An alternative to this would be: make the origin whatever the exact center of the page is. That one doesn’t need hotkeys and feels much more intuitive.

hluedeke:
Yeah…as I think about this…center of the page is what I think most people would expect to happen. It’s unlikely that people are cutting way off in the corners for most cases.

courtneypattison:
The rotation and flips happen from the centre of the PDF right now. My first thought was also to do it based on last click but I think people will feel like the transforms are buggy if they don’t understand where the origin is. The main use case for this one is unwrapping fold lines. It would remove the need for Inkscape for that feature. I feel like there has to be an elegant solution for this one but I can’t think of it.

roel-v:
I used it to unfold a fold line last weekend by cutting the normal shape, mirroring, moving the PDF on screen and also my fabric a bit until everything was where I wanted it to be. Is the problem here that a mirrored piece can end up far away from its original if it's off center on the page? If that's it, I think being able to set a mirror line manually is lot of difficult UI hassle for little gain. Using the current mouse position as the line when a keyboard key is pressed seems, while clever, pretty much undiscoverable to me, and it's not something one uses to often that it becomes ingrained. It also doesn't map to touch devices, which (much to my surprise) a majority of users seem to use (although that's only the impression I get from Facebook, stats may prove me wrong).

[...]

PR #127, opened 2024-03-22
Matrix based transforms

j-huff:
Hear me out...

In my opinion, all transformations applied to the PDF by tools (like rotate, flip, etc) should be applied to a single transformation matrix. The downside of this is that you are unable to completely disambiguate the transforms that are being applied to the PDF. This means you cannot determine things like its absolute mirror state (I could be mistaken about this, but from my testing it doesn't work). I've change the two state mirror buttons to single state. You *can* still determine things like it's position, rotation, and scale from the matrix.

This have a lot of benefits:
- Calculating the final transform is simple. Adding more tools doesn't change the calculation.
- Rotation button works as expected. Before, the pattern would be rotated anti-clockwise if it was mirrored.
- Adding more transformation tools will be much easier.
- If we implement history we most likely only need to track a single matrix

The TransformSettings.matrix is in unitsOfMeasure. This should make it straightforward to apply exact translations to it for future tools like grid snapping and "translate by".

A nice little bonus of this implementation is that the position of the PDF doesn't get reset if the calibration is modified, and recentering works properly now (it was a little off before).

Let me know what you think of this change.

### DIFF
--- a/app/[locale]/calibrate/page.tsx
+++ b/app/[locale]/calibrate/page.tsx
@@ -200,10 +200,10 @@ export default function Page() {
 
   useEffect(() => {
     /* Combine the translation portion of tranformSettings
-     * with the localTransformMatrix */
+     * with the localTransformMatrix. Note that the transformSettings
+     * matrix must be scaled by ptDensity first */
     const ptDensity = getPtDensity(unitOfMeasure);
     const translationMatrix = extractTranslationMatrix(Matrix.mul(transformSettings.matrix, ptDensity));
-    console.log(JSON.stringify(translationMatrix))
     const m = localTransform.mmul(translationMatrix);
     setMatrix3d(toMatrix3d(m.mmul(pdfTranslation)));
 
--- a/app/[locale]/calibrate/page.tsx
+++ b/app/[locale]/calibrate/page.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import { Matrix, inverse } from "ml-matrix";
-import { ChangeEvent, useCallback, useEffect, useState } from "react";
+import { ChangeEvent, useCallback, useEffect, useState, useRef} from "react";
 import { FullScreen, useFullScreenHandle } from "react-full-screen";
 
 import CalibrationCanvas from "@/_components/calibration-canvas";
@@ -12,7 +12,8 @@ import {
   getPerspectiveTransformFromPoints,
   toMatrix3d,
   decomposeTransformMatrix,
-  extractTranslationMatrix
+  scaleMatrixTranslation,
+  translate,
 } from "@/_lib/geometry";
 import { OverlayMode } from "@/_lib/drawing";
 import isValidPDF from "@/_lib/is-valid-pdf";
@@ -82,6 +83,9 @@ export default function Page() {
     horizontal: "",
     vertical: "",
   });
+
+  const pdfRef = useRef<HTMLDivElement | null>(null);
+  const [pdfDimensions, setPdfDimensions] = useState({ width: 0, height: 0 });
 
   function getDefaultPoints() {
     const o = 150;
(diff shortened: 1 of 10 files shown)


j-huff:
Adding the "mirror about line" tool we were discussing in #124 will be much easier with change.

courtneypattison:
I agree! I've been thinking about switching to a single matrix with transformations for a while. Thank you!

j-huff:
No, this shouldn't change anything about the calibration. The only major thing that has been changed is TransformSettings, which only gets applied to the position of the PDF. TransformSettings does get loaded from local storage, but from my understanding my modification shouldn't cause an error when migrating.

hluedeke:
> No, this doesn't change anything about the calibration. The only major thing that has been changed is TransformSettings, which only gets applied to the position of the PDF. TransformSettings does get loaded from local storage, but from my understanding my modification shouldn't cause an error when migrating.

Ok, great! I haven't really delved into the positioning or transformation code all that much, only enough to do simple things. So I appreciate the clarification on this.

> Edit: I'll test this just to be safe.

Appreciate it! I don't doubt you, but it doesn't hurt to be sure!

j-huff:
Something to note: transformSettings is a dependency of draw, so now that it gets updated when dragging the PDF, the canvas will get re-rendered every frame of the drag mouse movement. This may be undesirable from a performance standpoint, as canvas renders can be pretty expensive. The reason transformSettings is a dependency of the draw useEffect is because it also stores the current color mode. It might be a good idea to separate the matrix and color mode to prevent unneeded re-renders.

In the future, we may want to add tools that use the canvas for drawing, and need re-rendered when the pdf is moved. I still think that modifying the matrix should only trigger canvas rerenders if the active tool absolutely needs it, so for now I'm going to sever that dependency and we'll cross that bridge when we get to it.

hluedeke:
> Something to note: transformSettings is a dependency of draw, so now that it gets updated when dragging the PDF, the canvas will get re-rendered every frame of the drag mouse movement. This may be undesirable from a performance standpoint, as canvas renders can be pretty expensive. The reason transformSettings is a dependency of the draw useEffect is because it also stores the current color mode. It might be a good idea to separate the matrix and color mode to prevent unneeded re-renders.

Yeah, let's get rid of the dependency if it makes sense to do so. Preventing unnecessary re-renders is great! You can also make children of transformSettings the actual dependency (eg. transformSettings.color), instead of the whole transformSettings object. This might be the way to go if it doesn't otherwise make sense to break those up.

> In the future, we may want to add tools that use the canvas for drawing, and need re-rendered when the pdf is moved. I still think that modifying the matrix should only trigger canvas rerenders if the active tool absolutely needs it, so for now I'm going to sever that dependency and we'll cross that bridge when we get to it.

In that case, wouldn't we add the tool's state as a dependency in lieu of the actual matrix? You certainly wouldn't want both if they are linked. I'd say if you can remove extra dependencies and maintain current functionality, go for it.

j-huff:
To me, it made sense to just go ahead and separate transformSettings from displaySettings, and remove transformSettings as a dependency. I also combined the overlayMode with displaySettings, and am now storing overlayMode in localStorage with the other displaySettings.

j-huff:
> I just took a peek - do we need a whole object for TransformSettings now that it only has the matrix as the child? Is there a reason to hold onto that structure anymore?

Right now there isn't a good reason. The only reason I can think of is if in the future if we *really* wanted to have some other transform that get's treated separately from the matrix, although I don't know what that would be.

j-huff:
Ah, I goofed it when. Reflections are back to being weird.

j-huff:
Okay, doubled checked that calibration doesn't change on migration and it's fine. This branch should be good to go.

courtneypattison:
That centre screen rotation is so much better. Such a good idea @hluedeke.

I didn't realize how much the recenter being off centre bothered me until this PR. It's so much nicer now! Thank you @j-huff!

PR #135, opened 2024-03-23
Flip across center

j-huff:
As discussed in #124

I added a button to the toolbar the toggles "Flip across center" mode. Could probably work on the naming/tooltip but I'm a bit sleepy.

Functions as discussed: when enabled, the pdf is flipped across either the x-axis or y-axis. The flipping matrix code now has a "xintercept" or "yintercept" parameter for telling it what line to flip across, so adding "custom flip lines" should be easy later on if this doesn't suffice.


courtneypattison:
I'm a bit confused on this one lol! I think I did a bad job of explaining what I was thinking in #124. Don't we already flip about the centre from your last PR? I was thinking we hold off on adding the button, but just add the functionality to flip about the last mouse click with a couple of hotkeys, maybe 'h' and 'v' to see if we end up using them enough to add a button. What do you think?

courtneypattison:
> I have a feeling that even the click origin may not be needed anymore (there might be special cases).

I’m thinking the same thing. The centre screen origin feels so right. I think modifying the origin will likely frustrate more than delight.

courtneypattison:
We could maybe add the crosshair to the grid overlay options, then people could use that for unwrapping fold lines?

hluedeke:
> We could maybe add the crosshair to the grid overlay options, then people could use that for unwrapping fold lines?

I get what you're saying, but I don't know if I'd want to sacrifice my border or grid for a crosshair. This feels like an advanced setting. I'll start a discussion on advanced settings in the discussion tab.

j-huff:
> Don't we already flip about the centre from your last PR? 

I hope not. Is that what is happening for you? Because it shouldn't be lol.

courtneypattison:
***facepalm*** I didn’t move the pdf so the new default to centre made me think the transform origin moved. Sorry about that.

courtneypattison:
I’m thinking the flipping and rotating should always happen about the centre. Then maybe have the button to show the crosshair and potentially be able to move the crosshair in the future. What do you think? Did you see the discussion thread @hluedeke added for organizing the tools? I think we need to settle on the advanced toolbar stuff before adding in new buttons. I may pull the measuring tool out of beta until we decide how to organize them since that’s more of an advanced tool that a lot of people probably won’t use.

j-huff:
I think having "flip across center" always on would be fine for now, and having the centerlines be part of the overlay modes could work as well. If we do that then it might be best to change overlay "modes" to overlay "options", where each item is individually togglable in the drop-down (checkboxes), with a "disable" checkbox at the top of the list which hides all overlays (but doesn't uncheck your boxes in case you want to quickly toggle off and on with your previous selections still intact). This would allow the centerline overlay to be used with any combination of overlays and save on toolbar real estate.

courtneypattison:
I feel like that’s a simple solution and I like it. I’d like to know what @hluedeke thinks though. She’s so good at UI/UX stuff!

hluedeke:
I'm ok with trying it - can we separate out the border and grid so they are independently toggled? And are we saying that we can't move the flip/rotate lines? If that's the case, I think it'd be ok to add a centerline. If it was moveable, I'd have a harder time lumping it in with the other "static" options

courtneypattison:
> can we separate out the border and grid so they are independently toggled?

I think that’s exactly what @j-huff is proposing with the checkboxes. 

> And are we saying that we can't move the flip/rotate lines?

Yeah :)

j-huff:
Okay this branch should be good to go. Note that I changed the origin to the center, and drawing on the canvas is now done in unitsOfMeasure (you no longer need to multiply everything by ptDensity for it to look correct, 1 drawing unit = 1 inch/cm).

courtneypattison:
> Okay this branch should be good to go.

Awesome, thank you!

>  Note that I changed the origin to the center, and drawing on the canvas is now done in unitsOfMeasure (you no longer need to multiply everything by ptDensity for it to look correct, 1 drawing unit = 1 inch/cm).
> 

That’s a lot cleaner!

Do you think we should add rotating about the centre to this PR or do that elsewhere?

j-huff:
> Do you think we should add rotating about the centre to this PR or do that elsewhere?

I hadn't seen that discussed. Is that something people are requesting?

courtneypattison:
I believe we discussed centre screen rotation in #124. PP throwing large PDFs off screen now! Bad design choice on my part.

j-huff:
Addressed #141

Working on center screen rotation now

j-huff:
The PDF now rotates around the center :+1:

j-huff:
> PP is throwing large PDFs off screen now! 

I'm not sure what you mean by this

courtneypattison:
This looks so good! Thank you!

Since the rotation origin was the centre of the pdf, if you were at one end of a long pdf and rotated 90 degrees the whole thing would go off screen.

j-huff:
> Since the rotation origin was the centre of the pdf, if you were at one end of a long pdf and rotated 90 degrees the whole thing would go off screen.

Ahhh I see. I suppose this PR will prevent that as well.

courtneypattison:
Yeah, I’m excited about this one!

Issue #10, opened 2024-02-22
Add offline support


courtneypattison:

roel-v:
I just got an Electron version running which would allow a standalone/offline installer to be build:

![image](https://github.com/Pattern-Projector/pattern-projector/assets/568567/e380720b-e971-4ebb-9e3b-b5c2cf00855b)

This is using the newlook branch. It's using a development version of Electron, i.e. running from a dev server on localhost and not with a completely separate built .exe yet. From what I understand this shouldn't be too hard to add, it's only a matter of configuring things correctly, but until about 3 hours ago I didn't even know how to run a React app on my local machine so I'm just figuring things out as I go.

How do you envisage this? You want me to put all this together? For Windows would you prefer an MSI or NSIS installer? Do you care about Linux support at all? I don't know anything about Macs but there will probably be sewists with a Mac - Linux users will be rare in that population I'd think. I don't know yet if there is some way of cross compiling for Mac from Windows either. Ideas and/or direction welcome.

courtneypattison:
> I just got an Electron version running which would allow a standalone/offline installer to be build:
> 
[...]

PR #145, opened 2024-03-24
Adds offline support

courtneypattison:
Needs some love and testing but the basics are here for #10.


courtneypattison:
I'm happy enough with this one for now. It doesn't cache the gif from the first page and only caches the pages that you visit while online. There's definitely room for improvement but as long as this doesn't break anything I think it's better than no offline support. Thoughts?

hluedeke:
> I'm happy enough with this one for now. It doesn't cache the gif from the first page and only caches the pages that you visit while online. There's definitely room for improvement but as long as this doesn't break anything I think it's better than no offline support. Thoughts?

Sure 👍 I have learned that a simple download button was far more complicated than I expected and that not all platforms support PWA's (the concept seems to be a bit fringe). I don't know much about PWA's, but any support is probably better than none 🤷‍♀️

courtneypattison:
Lol, yeah, they definitely are fringe and may not be around that much longer! The way I've added offline support is with a service worker, so even if the site isn't "downloaded" it will be available offline in the browser, so going to https://www.patternprojector.com/ will still work when the device is offline.

hluedeke:
> Lol, yeah, they definitely are fringe and may not be around that much longer! The way I've added offline support is with a service worker, so even if the site isn't "downloaded" it will be available offline in the browser, so going to https://www.patternprojector.com/ will still work when the device is offline.

That works for me! I also want to throw out that maybe having a separate url for /calibrate and /project should consider your service worker implementation. It might be better to have fewer urls if you need to visit them all to have offline support (maybe you could implement the project page with a query variable instead, like: /calibrate?project=true or something like that)

courtneypattison:
> That works for me! I also want to throw out that maybe having a separate url for /calibrate and /project should consider your service worker implementation. It might be better to have fewer urls if you need to visit them all to have offline support (maybe you could implement the project page with a query variable instead, like: /calibrate?project=true or something like that)

Totally forgot to consider that issue! Good idea. Alternatively, it's possible to set up the service worker to precache specific pages, but I haven't learned how to do that yet!

PR #151, opened 2024-03-25
Added eslint rules to prevent tailing whitespaces, tabs, and incorrect indentation.

j-huff:
You can run yarn lint --fix to automatically fix most errors (except no-tabs, you can fix these errors by opening the file in vim and using :retab assuming your indentation is set up correctly).

How do y'all feel about adding more lint rules? I added some rules to let us get away with things like "any" and "Function". Unused vars are left as a warning, and this commit made no effort to fix them. If we decide that we like these lint rules, I'll go ahead and fix more of the unused vars warnings.

I mostly did this because I noticed that I've been using tabs instead of spaces because I messed up my vimrc on my home linux machine (oops!). This should help to keep the codebase a little cleaner (automagically).

There's apparently ways to set up your text editor to automatically run `yarn lint --fix` on save, which could be useful.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,3 +1,18 @@
 {
-  "extends": "next/core-web-vitals"
+  "extends": [
+    "next/core-web-vitals",
+    "plugin:@typescript-eslint/recommended"
+  ],
+  "parser": "@typescript-eslint/parser",
+  "plugins": [
+    "@typescript-eslint"
+  ],
+  "rules": {
+    "no-trailing-spaces": "error",
+    "@typescript-eslint/indent": ["error", 2],
+    "no-tabs": "error",
+    "@typescript-eslint/no-explicit-any": "off",
+    "@typescript-eslint/no-unused-vars":"warn",
+    "@typescript-eslint/ban-types":"off"
+  }
 }
--- a/app/[locale]/calibrate/page.tsx
+++ b/app/[locale]/calibrate/page.tsx
@@ -27,7 +27,7 @@ import {
 import {
   getDefaultDisplaySettings,
   DisplaySettings,
-	OverlaySettings
+  OverlaySettings
 } from "@/_lib/display-settings";
 import { CM, IN, getPtDensity } from "@/_lib/unit";
 import { Layer } from "@/_lib/layer";
@@ -109,23 +109,23 @@ export default function Page() {
 
   function resetTransformMatrix() {
     /* Resets and recenters the PDF */
-    let newTransformMatrix = Matrix.identity(3,3);
+    const newTransformMatrix = Matrix.identity(3,3);
     let tx = 0;
     let ty = 0;
 
     const scale = getPtDensity(unitOfMeasure) * .75;
-		const pdfWidth = layoutWidth / scale;
-		const pdfHeight = layoutHeight / scale;
-		
-		/* If pdf exceeds the width/height of the calibration
-			 align it to left/top */
-		if (pdfWidth > +width){
-			tx = (pdfWidth-(+width)) * 0.5;
-		}
-		if (pdfHeight > +height){
-			ty = (pdfHeight-(+height)) * 0.5;
-		}
-		
+    const pdfWidth = layoutWidth / scale;
+    const pdfHeight = layoutHeight / scale;
+
+    /* If pdf exceeds the width/height of the calibration
+       align it to left/top */
+    if (pdfWidth > +width){
+      tx = (pdfWidth-(+width)) * 0.5;
+    }
+    if (pdfHeight > +height){
+      ty = (pdfHeight-(+height)) * 0.5;
+    }
+
     const m = translate({ x: tx, y: ty});
     const recenteredMatrix = overrideTranslationFromMatrix(
       newTransformMatrix,
@@ -242,7 +242,7 @@ export default function Page() {
 
       const defaultDS = getDefaultDisplaySettings();
 
-      newDisplaySettings.overlay = localSettings.overlay !== undefined 
+      newDisplaySettings.overlay = localSettings.overlay !== undefined
         ? localSettings.overlay : defaultDS.overlay;
       newDisplaySettings.inverted = localSettings.inverted !== undefined
         ? localSettings.inverted : defaultDS.inverted;
@@ -286,7 +286,7 @@ export default function Page() {
   /* Update the pdfWidth and pdfHeight when it changes */
   useEffect(() => {
     const observer = new ResizeObserver((entries) => {
-      for (let entry of entries) {
+      for (const entry of entries) {
         const { width, height } = entry.contentRect;
         /* Only trigger if the dimension is non-zero
          * and different that the current dimension */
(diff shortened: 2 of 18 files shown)


hluedeke:
Are you using an IDE or are you using vim to program? There is a package in the project (should be installed in your node_modules folder) called "prettier" that you can also run that fix spacing. You can do it through the command line, or if you are using an IDE, you can set it up to automatically run on file save. Linting is great (and fine with me if we want to implement in builds), but prettier will solve the spacing issues!

hluedeke:
> Are you using an IDE or are you using vim to program? There is a package in the project (should be installed in your node_modules folder) called "prettier" that you can also run that fix spacing. You can do it through the command line, or if you are using an IDE, you can set it up to automatically run on file save. Linting is great (and fine with me if we want to implement in builds), but prettier will solve the spacing issues!

I also want to throw out there that prettier and eslint can sort of compete with each other, so you have to be careful on implementation. It's not super difficult, but you have to take extra steps to be sure the correct rules are setup. I would say I prefer prettier for spacing issues and eslint for more semantic rules.

j-huff:
Didn't realize we were already using prettier 😅 I was looking into that as well but decided it was too strict. Removing PR.

courtneypattison:
I think some standard style is a great idea! I haphazardly added and eslintrc and prettier extension to keep things clean, then when Hannah started contributing my prettier extension kept muddying up PRs, so I set it to only run on modifications I make to PRs. I'd love it if we all used the same linter and prettier! Just not sure the best way to set it up.

hluedeke:
> I think some standard style is a great idea! I haphazardly added and eslintrc and prettier extension to keep things clean, then when Hannah started contributing my prettier extension kept muddying up PRs, so I set it to only run on modifications I make to PRs. I'd love it if we all used the same linter and prettier! Just not sure the best way to set it up.

I suspect we are different prettier versions. I think you had globally installed prettier, and I ended up adding a local installation to the project. I know what you mean about it muddying up PR's, but I think enforcing the same version by using a project-specific copy will fix that (at least I hope).

Airbnb has a published set of eslint rules they maintain (though it has changed over time) that I've used before. There are a lot of rules in there, haha. But whatever we want to do, I'm good with! I am used to having a linter bark at me all the time, so I think I just program that way by default now 😅

I don't have a particularly strong opinion on styling, as long as it doesn't take too much away from actual development time (which is why I love the prettier on save workflow).

j-huff:
Agree.  I do like the idea of having built into the lint rather than relying on people running prettier separately. Causing the build to fail due to poorly formatted code would prevent merging PRs with poor formatting. I am dubious of using prettier in the linter due to its sctrictness and potential conflict issues, but am still open to it. I think we could add enough typescript lint rules to get pretty close to what prettier does.

j-huff:
From my understanding, Prettier is not very configurable because the coding style it uses is already opinionated. Adding our own lint rules (possibly based on the rules by something like AirBnB) would let us employ our style in a more opaque manner.

hluedeke:
> Agree. I do like the idea of having built into the lint rather than relying on people running prettier separately. Causing the build to fail due to poorly formatted code would prevent merging PRs with poor formatting. I am dubious of using prettier in the linter due to its sctrictness and potential conflict issues, but am still open to it. I think we could add enough typescript lint rules to get pretty close to what prettier does.

I actually find it the opposite. Eslint (at least with airbnb rules applied) is far stricter than prettier is, and as you mentioned fails the build. To me, that's an argument for using it on semantic rules vs. spacing rules. And I have not found that eslint is capable of auto-fixing many of the rules that it complains about (it just complains - of course, maybe I'm using it wrong!), so I much prefer the auto-reformatting of prettier. Prettier doesn't/shouldn't do any updates to the code that would cause the code to be changed in significant ways, just indentation and such, which I consider to be less important than semantic issues that eslint would catch.

hluedeke:
> Causing the build to fail due to poorly formatted code (such as my tabs instead of spaces 😅) would prevent merging PRs with poor formatting

I think either package (or both packages) could be run automatically by setting them up in package.json. We just have to set the "build" script (or possibly even the "dev" script) to use these packages. Then they would just run automatically without worrying about whether or not individual devs ran them or not.

j-huff:
> I actually find it the opposite. Eslint (at least with airbnb rules applied) is far stricter than prettier is, and as you mentioned fails the build. To me, that's an argument for using it on semantic rules vs. spacing rules.

I suppose what I really meant "inflexible" rather than strict. Like with our own eslint rules we can make it as strict as we want, whereas the prettier in some cases cannot be configured to match the style that is already being used most places in the the project.

And as far as semantic rules goes, so far the only significant semantic rule that is employed by this minimal eslint config is that it changes var to let in every case (good thing IMO), and changes let to const when the variable doesn't get modified. Those seem pretty fair to me.

hluedeke:
> I suppose what I really meant is "inflexible" rather than strict. Like with our own eslint rules we can make it as strict as we want, whereas the prettier in some cases cannot be configured to match the style that is already being used most places in the the project.

I hear what you're saying. I think that's also where I meant that "I don't have a strong opinion on style". I think most places in the project are currently using prettier (although, admittedly, I think different versions or rulesets, which I do think we should figure out).

> And as far as semantic rules goes, so far the only significant semantic rule that is employed by this minimal eslint config is that it changes var to let in every case (good thing IMO), and changes let to const when the variable doesn't get modified. Those seem pretty fair to me.

Yeah! I'm not opposed to using a linter. I'm used to using one! There are many, many rules you can apply semantically to a linter, and many of them are very useful. I am definitely in favor of linting for semantics.

j-huff:
Looks like most "style" eslint rules are deprecated and replaced by esline-stylistic https://eslint.style/

j-huff:
I'm going to table this for now and just use prettier locally

courtneypattison:
I really like the idea of adding more code quality checks to the linter. The eslintrc isn't doing much for us as is. What about adding the rules from this PR and maybe adding [eslint-config-prettier](https://github.com/prettier/eslint-config-prettier) to avoid conflicting styles and we can all run the same version of prettier locally to get the style fixes on save? Deal with code formatting with prettier and code quality with eslint?

hluedeke:
> I really like the idea of adding more code quality checks to the linter. The eslintrc isn't doing much for us as is. What about adding the rules from this PR and maybe adding [eslint-config-prettier](https://github.com/prettier/eslint-config-prettier) to avoid conflicting styles and we can all run the same version of prettier locally to get the style fixes on save? Deal with code formatting with prettier and code quality with eslint?

This is my vote!

hluedeke:
Here's an example of the airbnb rules. There are a ton, and I'm not necessarily saying we go with this package, but it will give you an idea of the kinds of options we have.

https://github.com/airbnb/javascript/tree/master/packages/eslint-config-airbnb-base/rules

courtneypattison:
> This is my vote, minus these rules:
> 
> ```
> "no-trailing-spaces": "error",
>     "@typescript-eslint/indent": ["error", 2],
>     "no-tabs": "error"
> ```
> 
> Those are style-specific and not semantic.
> 
> === EDIT AGAIN
> 
> Sorry if that last bit came across a bit harsh. The prettier eslint rule will invalidate any style rules that it overwrites, so we can leave these in if you guys want!

You're not harsh at all! I think this combination of your ideas will clean the code up considerably. What do you think @j-huff? I agree that those lines can stay or go, no biggie.

PR #152, opened 2024-03-25
Prepare for when the French translation is done - this adds support for French to the rest of the app.

roel-v:
Also puts Danish as 'Dansk' into the language menu, to be consistent with the others and the idea behind how the drop down should work.

### DIFF
--- a/app/[locale]/page.tsx
+++ b/app/[locale]/page.tsx
@@ -20,7 +20,7 @@ import { IconButton } from "@/_components/buttons/icon-button";
 export default function Home() {
   const t = useTranslations("HomePage");
   const messages = useMessages() as IntlMessages;
-  const keys = Object.keys(messages.HomePage.resources.links);
+  const keys = Object.keys(messages.HomePage.resources.links ? messages.HomePage.resources.links : {});
 
   return (
     <main className="m-4">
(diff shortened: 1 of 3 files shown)


courtneypattison:
Thank you! I thought about doing this and then didn’t write it down and promptly forgot. I messaged the person who started the French translation to thank them and they’re said they’ll complete the whole translation for us. Woot woot!

roel-v:
There is one more issue, if there are no additional resources in the messages/xx.json file, that whole section should not be shown on the main page. I can do a crude 'put the whole section into a string and make that string empty when there's no 'links' element found', but that would mess up the flow of the html in the jsx, there's probably a more idiomatic way to do what is essentially an 'if' in a case like this? Chatgpt is down for me so I'm flying blind lol.

courtneypattison:
I’m not sure! Im only on my phone right now so can’t be much help. I can look into it later though.

BErika:
What about using English when there is no translation for some key? Franglish would be better that missing information, isn't it?

roel-v:
Yes if that's the term that is most commonly used. E.g. I used 'Overlay' in the Dutch translation as-is, that's the term that is usually used in Dutch for that concept. But yes all strings have to be present in the translation.

BErika:
My suggestion was not clear sorry. I am talking about strings missing in the translation.

I have 2 use cases in mind:
1. strings I don't know how to translate 
2. keys that will be added in the future but the translation might not be kept up to date so there will be no translation for them

Currently, from what I have experienced, we either display nothing if this is a hover, or we print the key. I suggest that when there is no translation available, we use the English strings. That would result in a mix of French (or other language) and English but I think it is better to have English information in the middle of French text instead of missing information.

courtneypattison:
Totally agree! Seems to be easy enough to add with [deepmerge](https://next-intl-docs.vercel.app/docs/usage/configuration).

courtneypattison:
Merged into #158 :)

PR #153, opened 2024-03-25
Overlay Opacity

j-huff:
Adds an overlay opacity number input to the overlay dropdown menu. Other minor formatting fixes.

Thinking of changing the number input to a slider but that seemed a bit too much for my current laziness. 

Edit: Also fixes the react warning complaining about the checkbox not having onChange or readOnly set.

### DIFF
--- a/app/[locale]/calibrate/page.tsx
+++ b/app/[locale]/calibrate/page.tsx
@@ -27,7 +27,7 @@ import {
 import {
   getDefaultDisplaySettings,
   DisplaySettings,
-	OverlaySettings
+  OverlaySettings
 } from "@/_lib/display-settings";
 import { CM, IN, getPtDensity } from "@/_lib/unit";
 import { Layer } from "@/_lib/layer";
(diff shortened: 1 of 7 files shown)


courtneypattison:
I'm a bit concerned about this option adding too much complexity for people to parse when they open that overlay menu. What is the issue we're trying to solve with this one? Do you find the lines too dark? Could we make them all dashed lines, thinner, a different opacity, or something like that?

j-huff:
I'll get rid of the opacity menu item, but this PR at least fixes the checkbox warning

j-huff:
whoops

courtneypattison:
Cool beans, thank you! I'm on a bit of a mission to remove unused code today. Do you think you could remove any of the opacity logic, strings, types, etc pretty please? This code base is getting unwieldily! Also, what is 'Enter' being used for here?

j-huff:
> Also, what is 'Enter' being used for here?

I was making sure the menu was usable via tab/shift+tab and enter. I forgot to take out the onKeyDown callback though, as it's not needed with the htmlFor tag.

j-huff:
Forgot to remove translations for opacity.

courtneypattison:
Hmm, which part of this fixes the react warning? Could we simplify it just down to that? It still seems like it’s adding complexity that isn't used. I’m only looking at the code on my phone so I could be misinterpreting it. It’s so hard to read on a tiny screen!

j-huff:
Just to note, most of the changes outside of dropdown-checkbox-icon-button were just switching tabs to spaces, so this may look like a bigger change than it actually is.

As far as added complexity in dropdown-checkbox-icon-button...

Yes, this does now have the ability to have a numeric option in the menu, which is not being used. However, I think don't think this really adds too much complexity to the component. I could still remove it if you prefer.

There are other good changes to the component IMO, such as use of generics to ensure that the keys in optionSettings are actually in options. Additionally, the menu items are no longer buttons, instead using the more intuitive "label" with a htmlFor tag. My understanding is that this is better for accessibility.

courtneypattison:
> Edit: I could make a new branch that only has the changes to the dropdown button component rather than being cluttered with the tabs/spaces nonsense.

That would be amazing, thank you! I'm so bad at merge conflicts. Sorry for taking so long to respond. I've been trying to focus on getting a video made for this release and I have no video making/editing skills so it's slow going.

PR #154, opened 2024-03-26
Added shift key modifier to arrow key movement

j-huff:
Holding shift while pressing arrow keys multiplies the movement by 16 (default).

For moving the PDF, the holding shift makes the arrow keys move the PDF by 1 inch/cm. Without the shift key it's moved by 1/scale (scale = 16, so 1/16th inch/cm).

### DIFF
--- a/app/[locale]/calibrate/page.tsx
+++ b/app/[locale]/calibrate/page.tsx
@@ -80,6 +80,7 @@ export default function Page() {
   const [layoutHeight, setLayoutHeight] = useState<number>(0);
   const [lineThickness, setLineThickness] = useState<number>(0);
   const [measuring, setMeasuring] = useState<boolean>(false);
+  const [gridSize, setGridSize] = useState<number>(16.0);
 
   const [showStitchMenu, setShowStitchMenu] = useState<boolean>(false);
   const [pageRange, setPageRange] = useState<string>("");
@@ -89,6 +90,7 @@ export default function Page() {
     vertical: "",
   });
 
+  /* Needed for final perspective transform */
   const pdfRef = useRef<HTMLDivElement | null>(null);
   const [pdfDimensions, setPdfDimensions] = useState({ width: 0, height: 0 });
 
@@ -255,8 +257,8 @@ export default function Page() {
     }
   }, []);
 
-  /* Scale of 1.0 would mean 1 in/cm per key press. Here it is 1/16th in or  1 mm */
-  const arrowKeyScale = unitOfMeasure == IN ? 16.0 : 10.0;
+  /* Scale of 1.0 would mean 1 in/cm per key press. 16.0 -> 1/16th in/cm */
+  const arrowKeyScale = gridSize;
   useProgArrowKeyToMatrix(!isCalibrating, arrowKeyScale, (matrix) => {
     const newTransformMatrix = matrix.mmul(transformSettings.matrix);
     setTransformSettings({
@@ -264,6 +266,10 @@ export default function Page() {
       matrix: newTransformMatrix,
     });
   });
+
+  useEffect(() => {
+    setGridSize(unitOfMeasure === IN ? 16.0 : 10.0);
+  }, [unitOfMeasure])
 
   useEffect(() => {
     /* Combine the translation portion of tranformSettings
@@ -317,7 +323,8 @@ export default function Page() {
     const w = Number(width);
     const h = Number(height);
     if (points && points.length === maxPoints) {
-      let m = getPerspectiveTransformFromPoints(points, w, h, ptDensity, true);
+      /* In units of measure */
+      let m = getPerspectiveTransformFromPoints(points, w, h, 1.0, true);
       let n = getPerspectiveTransformFromPoints(points, w, h, ptDensity, false);
       setPerspective(m);
       setLocalTransform(n);
@@ -391,11 +398,8 @@ export default function Page() {
               }
             }}
             pageCount={pageCount}
-            localTransform={localTransform}
-            setLocalTransform={setLocalTransform}
             layoutWidth={layoutWidth}
             layoutHeight={layoutHeight}
-            calibrationTransform={calibrationTransform}
             lineThickness={lineThickness}
             setLineThickness={setLineThickness}
             setShowStitchMenu={setShowStitchMenu}
@@ -467,6 +471,7 @@ export default function Page() {
             setTransformSettings={setTransformSettings}
             perspective={perspective}
             unitOfMeasure={unitOfMeasure}
+            gridSize={gridSize}
           >
             <div
               ref={pdfRef}
--- a/app/[locale]/calibrate/page.tsx
+++ b/app/[locale]/calibrate/page.tsx
@@ -254,8 +254,9 @@ export default function Page() {
     }
   }, []);
 
-  /* Scale of 1.0 would mean 1 in/cm per key press. Here it is 1/16th in/cm */
-  useProgArrowKeyToMatrix(!isCalibrating, 16.0, (matrix) => {
+  /* Scale of 1.0 would mean 1 in/cm per key press. Here it is 1/16th in or  1 mm */
+  const arrowKeyScale = unitOfMeasure == IN ? 16.0 : 10.0;
+  useProgArrowKeyToMatrix(!isCalibrating, arrowKeyScale, (matrix) => {
     const newTransformMatrix = matrix.mmul(transformSettings.matrix);
     setTransformSettings({
       ...transformSettings,
--- a/app/[locale]/calibrate/page.tsx
+++ b/app/[locale]/calibrate/page.tsx
@@ -81,6 +81,7 @@ export default function Page() {
   const [lineThickness, setLineThickness] = useState<number>(0);
   const [measuring, setMeasuring] = useState<boolean>(false);
   const [gridSize, setGridSize] = useState<number>(16.0);
+  const [isDragging, setIsDragging] = useState<boolean>(false);
 
   const [showStitchMenu, setShowStitchMenu] = useState<boolean>(false);
   const [pageRange, setPageRange] = useState<string>("");
@@ -412,6 +413,7 @@ export default function Page() {
             matrix={transformSettings.matrix}
             unitOfMeasure={unitOfMeasure}
             className={`z-30 ${visible(!isCalibrating)}`}
+            preventReset={isDragging}
           />
 
           <LayerMenu
@@ -472,6 +474,7 @@ export default function Page() {
             perspective={perspective}
             unitOfMeasure={unitOfMeasure}
             gridSize={gridSize}
+            setIsDragging={setIsDragging}
           >
             <div
               ref={pdfRef}
--- a/app/[locale]/calibrate/page.tsx
+++ b/app/[locale]/calibrate/page.tsx
@@ -8,6 +8,7 @@ import CalibrationCanvas from "@/_components/calibration-canvas";
 import Draggable from "@/_components/draggable";
 import Header from "@/_components/header";
 import PDFViewer from "@/_components/pdf-viewer";
+import TransformMatrixDisplay from "@/_components/transform-matrix-display";
 import {
   getPerspectiveTransformFromPoints,
   toMatrix3d,
@@ -27,7 +28,7 @@ import {
 import {
   getDefaultDisplaySettings,
   DisplaySettings,
-	OverlaySettings
+  OverlaySettings
 } from "@/_lib/display-settings";
 import { CM, IN, getPtDensity } from "@/_lib/unit";
 import { Layer } from "@/_lib/layer";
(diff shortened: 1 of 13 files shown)


roel-v:
Can this be made so that when the unit is metric, it moves by 1 mm on each key press? 1/16th of a cm is (in metric) a useless amount (I understand why it is the way it is, and I'm not sure how difficult it is to keep things consistent app-wide when this sort of conversions start to diverge between imperial/metric usage)

j-huff:
It's only useless until metric users learn about fractions 😉

But yes I'll add that tomorrow

roel-v:
Well that's a discussion for some other time 😉but yeah I'm sure metric using PP users will love this. Can we think of a way to communicate to the user that they're moving in useful real-world amounts? I think most users (like myself) are used to positioning and moving things in units that don't have direct real-world meaning, which is very different here - if you know you're moving in real-world units, you can essentially e.g. extend a pair of pants without even opening an editing program or even using your tape measure.

I don't have any examples in mind of software that gives this feedback directly. Maybe show a text in the corner of how much the pattern has been moved? But since when? Since the last time the pattern was moved with the mouse? (I'm high jacking this PR here I guess, maybe open another issue is better)

j-huff:
> Maybe show a text in the corner of how much the pattern has been moved?

Yeah, that's something I've been considering for a while actually.

courtneypattison:
I love this! Thank you! I do think we should show the user what's going on in some way, maybe in a future release, but I'm not sure what would be the best way to do that. If it were just in one dimension we could do something like how the measure tool works and hav a line with measurement draw as the movement is happening. But it's in two dimensions so that doesn't work.

One tiny thing: could you please switch to using the new useKeyDown hook? I'd like to start switching to that with any new key handling PRs.

j-huff:
Yep I'll switch it over. Also working on a position display right now, almost finished. I think it's unobtrusive enough to just throw in this PR.

courtneypattison:
Great, thank you! Another tiny nit is to consider using the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) when setting based on whether something is undefined.

j-huff:
> Great, thank you! Another tiny nit is to consider using the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) when setting based on whether something is undefined.

Thanks, didn't know about that.

For the useKeyDown hook... Should I also add a useKeyUp hook? I need to track whether shift is being held, so I need both up and down.

courtneypattison:
If you need to do keyup please add a hook for it. You can see how I did shifting for the measuring tool #117 to see a different method!

hluedeke:
I want to comment on this quick, as I'm working on something similar with the auto-hide of the menus and the mouse button that you already wrote @j-huff and were having issues with the maximum stack call exceeding. If it's a document or window-level listener, I think it's best to throw it into a hook, as it's easier to see high-level events grouped in the same area in code. I also think that you might consider writing a second hook that handles the complete functionality of your shift button instead of trying to put it on the component itself. The goal is to keep listener functions and side effects abstracted from the normal flow of the page logic, if that makes sense. I'm currently working on (which is generous because I haven't had a lot of time) moving the auto-hide of the mouse button and the menus to a "hide-all-the-things-after-a-timeout-in-full-screen" hook that manages all those states in the same place. You might consider something similar for the shift key.

courtneypattison:
Thank you @hluedeke! I never know when to use a hook. This explanation really helps :)

hluedeke:
> Thank you @hluedeke! I never know when to use a hook. This explanation really helps :)

Haha, well, this is only my opinion! I'm sure people use hooks in very different ways, as they are literally just abstractions of state management (eg. useState, useEffect). I just have experience using key handlers, and I know that once you start down that road, collisions inevitably happen. It's very difficult to determine why your code isn't working when the handler is declared directly on the document from a component somewhere buried 5 levels down in your component tree :)

j-huff:
It might be advisable to add an optional key code parameter to the useKeyUp/Down hook callbacks to avoid having to make separate useKeyDown hooks for each arrow key like I did.

hluedeke:
> It might be advisable to add an optional key code parameter to the useKeyUp/Down hook callbacks to avoid having to make separate useKeyDown hooks for each arrow key like I did.

Yeah, go for it!

j-huff:
Okay this branch should be good to go.

j-huff:
I'm somewhat dubious of my useKeyHeld hook but it seems to work ¯\_(ツ)_/¯

courtneypattison:
This is looking great, thank you! I feel like some people will see delta and feel confused/overwhelmed. Any ideas for something a little simpler to show users? Maybe something like top, right, bottom, left showing how far they've moved the PDF in those directions?

j-huff:
Since when is it unreasonable to expect users to understand Greek alphabet math notation?  🤔

I'll think of something. I'm also going to add drag grid snapping to this PR as it seems natural since the position gets displayed as a fraction only if you are on the grid.

j-huff:
> Any ideas for something a little simpler to show users? Maybe something like top, right, bottom, left showing how far they've moved the PDF in those directions?

I want to show both the absolute position and the amount moved. Showing x, y and Δx,Δy is the simplest way to achieve that. Is you concern that people will not understand the delta symbol? Or that should both position and delta is confusing?

If we have to pick just one thing to show, I think delta would be more important. But if we just show it as x, y then wouldn't that possibly get confused for position? I guess I could put a label above it that says "Current movement:" or something like that.

I don't really get the meaning of "top, right, bottom, left" here. Top/bottom right/left would just be the negative of each other?

hluedeke:
@j-huff, I think it makes sense to show something like this:

Left: 1/2", Down: 1/4"

Have that change depending on where they've shifted. A lot of people using this site are not tech savvy (or possibly math savvy), so I would shy away from coordinates or mathematical symbols in general.

Also @roel-v we would need to translate left/down/up/right, I assume? Maybe we could use arrow icons? Like right arrow icon for +x, left arrow icon for -x, etc?

hluedeke:
> Only show the position in the little popup, but draw an arrow on the canvas from the start of the movement to the current position, with text indicating its magnitude. Similar to the measure tool but smaller.

This is more along the lines of visual representation that is more intuitive and a smaller learning curve. I wouldn't even bother with position (which could be confusing). I think we just need to know relative movement when in precision mode.

j-huff:
Okay I think this is more in line with what y'all are getting at.

courtneypattison:
That looks much simpler, thank you! What does the grid snapping do/what is it for?

j-huff:
> That looks much simpler, thank you! What does the grid snapping do/what is it for?

Let's you move the pattern by 1/16th inch increments using the mouse.

roel-v:
> Also @roel-v we would need to translate left/down/up/right, I assume? Maybe we could use arrow icons? Like right arrow icon for +x, left arrow icon for -x, etc?

Not sure if this comment is still relevant, things move too fast for me :) but if the translation is the reason for using arrows then I'd say that should not be a reason for choosing icons over text; the translation is a minor thing at this point.

I did try out the movement and while I like it, I'm not sure about the timeout and about the concept of the move distance being lost after some time. I feel very rushed and put under pressure to get it right the first time, once I even know what's happening, which I'm sure will not be understood by many of our users. I understand why this PR is like it is though, and I don't have an alternative proposal that I fully stand behind, either. The only other option is making the movement 'permanent' since some user-picked point in time. I.e., user would have to click a button that says 'make this the real position of the pattern' and all movement after that is displayed as distance from that user-defined point. But I don't like this solution because how do you communicate to the user how this works?

Overall, I'm thinking we are approaching a point where there is so much functionality, not all of it discoverable, that more documentation (and with it, a slower and more strictly managed release process, so that the docs aren't always playing catch up with the functionality) is becoming necessary.

j-huff:
I think the feeling of being rushed occurs only if you don't know how far you want to move the pattern ahead of time. I think the idea here was to give a way for people to move a pattern exactly 3/4 inch (Also it was to give the user feedback that the arrow keys move a meaningful amount per press as you suggested). If you are eyeballing the movement then does it really matter if the numbers disappear? You likely weren't looking at them to begin with.

Maybe for the use case of "slowly nudging a pattern over until it looks right and then checking to see how far you moved it", it would make sense to have the movement numbers be more persistent. It's unclear to me whether that is a worthwhile use case to cover.

roel-v:
Yes you're right, and indeed if you do know that, it's great. And from a software power user perspective it's great, it moves more towards CAD and a high efficiency/keyboard based exact measurements workflow, which I naturally gravitate towards anyway. When I sew though, it's very different (at least for me) and I try to evaluate things from that different perspective. Like, I totally see myself nudging something over by 1cm, taking out my ruler and put it on a piece of reference fabric to double check, and then change that 1cm to 2 or whatever. Or moving something over over to match a piece it needs to have the same length as, and then measuring several times IRL to make sure I have everything right in my head and on the cutting table. But as I said I don't really have a solution or proposal. As you say, it depends on the delineation of the use case.

j-huff:
It would actually be very easy to just chuck it in as a keybind. Hit L for lock and the numbers won't fade out. There's already a mechanism for doing this to keep it from fading out during a mouse drag. Very tempting to grow the list of secret keybinds.

courtneypattison:
> I did try out the movement and while I like it, I'm not sure about the timeout and about the concept of the move distance being lost after some time. I feel very rushed and put under pressure to get it right the first time, once I even know what's happening, which I'm sure will not be understood by many of our users. I understand why this PR is like it is though, and I don't have an alternative proposal that I fully stand behind, either. The only other option is making the movement 'permanent' since some user-picked point in time. I.e., user would have to click a button that says 'make this the real position of the pattern' and all movement after that is displayed as distance from that user-defined point. But I don't like this solution because how do you communicate to the user how this works?

I completely agree. There needs to be some permanence without cluttering up the already busy UI. What do you folks think about having the movement pop up when shifting and stays there until the user dismisses it with an x in the corner? I’m also concerned about the minimum size of movements. How small do people need to be moving? Is half a cm small enough? Showing fractions scares me too. I feel like people will see jumping fractions and be intimidated. 

> 
> Overall, I'm thinking we are approaching a point where there is so much functionality, not all of it discoverable, that more documentation (and with it, a slower and more strictly managed release process, so that the docs aren't always playing catch up with the functionality) is becoming necessary.


Yes! I’m planning on recording a video going over the functionality in beta, then merging into main tonight. I’m only going to have time once the kids go to sleep though so we’ll see what I can get done. I set things up so I can just run through things without editing the video so hopefully it doesn’t take too much time.

I’d also like to prioritize cleaning up the code and documentation next week.

j-huff:
> What do you folks think about having the movement pop up when shifting and stays there until the user dismisses it with an x in the corner? 

By "shifting" do you mean moving the pattern with arrow keys? I suppose that could work. More intuitive than a keybind, but also slightly more intrusive. I imagine most people just move the pattern with the mouse, so maybe it wouldn't be too much of a nuisance to need to close the popup any time you move with the arrow keys.

hluedeke:
I'll be honest, I haven't tried this yet, so my entire context is this conversation. But doesn't it make sense that if the user uses shift+arrow, that would trigger visibility of the "delta" movement, and then any action that drags the pdf not using shift+arrow would cause it to hide? (Dragging with the mouse, dragging with normal arrow keys)?

j-huff:
> I'll be honest, I haven't tried this yet, so my entire context is this conversation. But doesn't it make sense that if the user uses shift+arrow, that would trigger visibility of the "delta" movement, and then any action that drags the pdf not using shift+arrow would cause it to hide? (Dragging with the mouse, dragging with normal arrow keys)?

Shift + arrow just makes the movement bigger. The delta should still be displayed with small movements.

I could see just not displaying the delta for mouse movements, but requiring the user to move the pdf again to exit the overlay seems like it could be a nuisance. Like you get it to the perfect position with arrow keys, then to get rid of the delta display you need to move the pattern? Doesn't seem right.

courtneypattison:
> > I'll be honest, I haven't tried this yet, so my entire context is this conversation. But doesn't it make sense that if the user uses shift+arrow, that would trigger visibility of the "delta" movement, and then any action that drags the pdf not using shift+arrow would cause it to hide? (Dragging with the mouse, dragging with normal arrow keys)?
> 

I really like this idea, simple and elegant.

> Shift + arrow just makes the movement bigger. The delta should still be displayed with small movements.

I think only showing the popup only when shifting makes sense because it feels like an intentional activation of a special mode. I think I'd find the popup annoying if I were just using the arrow keys to move the PDF around.

> 
> I could see just not displaying the delta for mouse movements, but requiring the user to move the pdf again to exit the overlay seems like it could be a nuisance. Like you get it to the perfect position with arrow keys, then to get rid of the delta display you need to move the pattern? Doesn't seem right.

I'm not sure I agree on this one. Once you've moved the position with the mouse, the information should be lost because it is no longer valid.

j-huff:
> Once you've moved the position with the mouse, the information should be lost because it is no longer valid.

I wasn't necessarily disputing this, I was just saying that shouldn't be the only way that the popup disappears.

> I think I'd find the popup annoying if I were just using the arrow keys to move the PDF around.

I would find it annoying too, but only if it didn't disappear when I was done moving it, and I had to click a button to close it every time or something. In it's current state, I think there is good balance between persistence and unobtrusiveness, but only if people know the keybind (L) for making it persistent.

Considering how difficult it apparently is to workshop a solution to a problem that I created out of think air, maybe it's best to just scrap it.

courtneypattison:
Definitely don't scrap it! It's a great solution to moving pieces for adjustments. It's just tricky trying to balance making it obvious how to use it without unintentionally annoying people. I'd love for it to be a feature that anyone can use rather than having it be more of a power user feature because it's a common method of alteration I believe.

j-huff:
apologies, I tend to get melodramatic before my coffee kicks in 😅

courtneypattison:
> apologies, I tend to get melodramatic before my coffee kicks in 😅
> 

Don’t worry about it! I’ve been full melodrama since PP launched. 

> Edit: I'm likely going to make a new branch/pull request that will make this easier to merge.
Thank you!

courtneypattison:
Screenshot 2024-04-02 at 2 08 39 PM

This data is for people on the calibrate route. Definitely a lot more desktop users but enough tablet users to warrant designing for them as well. I'm working on an overhaul of the calibration page based on feedback from people in the group. I'm adding in a button based movement pad (arrow key pad sort of thing with the button in the middle for going to the next point) that can be toggled on and off for adjusting the corners. That's not ready for review yet though!

I was thinking maybe that movement pad could be repurposed for the projection page and the button in the middle could be replaced with something like an inline input where people could put how much they want to move with each press. Then we could have a sticky header for the movement pad that shows the current movement, like what is in this PR. Just throwing that out there. I've got to finish the calibration version to see if that even makes sense first.

Issue #150, opened 2024-03-25
Stitch menu resets PDF location with each setting change


courtneypattison:
Now that the transform origin in set to center, the stitch menu resets the location of the pdf whenever there is a setting change.

courtneypattison:
@j-huff what do you think about this one? I’m thinking about switching the transform origin back to the top left. The switch menu feels quite frustrating now.

j-huff:
Right, it's set to recenter whenever the layout changes. So if you go from show 1 page to 2 pages it recenters, or if you change the number of columns. I can remove that if it's annoying.

j-huff:
Since when is it unreasonable to expect users to understand greek alphabet math notation? 🤔

I'll think of something. Also I'm going to add drag grid snapping to this PR as it seems natural as the position gets displayed as fractions only if you are on the grid.

PR #156, opened 2024-03-27
Less reset center

j-huff:
fix for #150

### DIFF
--- a/app/[locale]/calibrate/page.tsx
+++ b/app/[locale]/calibrate/page.tsx
@@ -172,6 +172,7 @@ export default function Page() {
     const { files } = e.target;
 
     if (files && files[0] && isValidPDF(files[0])) {
+      resetTransformMatrix();
       setFile(files[0]);
     }
   }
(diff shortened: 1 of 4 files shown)


courtneypattison:
This is definitely better, but it still feels like a downgrade from how the stitch menu works on the main branch. I started refactoring the geometry library to see what I can do about this issue. I don't have anything ready enough to share yet though!

j-huff:
I don't understand. What exactly is different?

j-huff:
@courtneypattison I can tell you that whatever issue you are experiencing, it is most likely not due to the geometry library.

It is extremely difficult to have the following features co-exist:
- Align to the left on long document load, align to center if the document fits in the grid
- Document transform matrix is relative to the recenter of the document

Basically, it's difficult to trigger the recenter function once a new document is loaded but only once the layout states are completely finished being set. I was able to get away away with the left alignment by just calling resetTransformMatrix any time the layoutWidth or layoutHeight changed.

So possible solutions are:

- stop trying to align the PDF to the left on load
- make the document transform matrix origin be the upper left of the pdf (extremely annoying to do matrix operations if we go this route)
- Figure out a way of triggering the resetTransformMatrix after layoutWidth and layoutHeight has finished being updated ONLY by a new document being loaded. This one seems simple on the surface but is actually difficult. Maybe not impossible though?


Take your pick I guess.

j-huff:
> Figure out a way of triggering the resetTransformMatrix after layoutWidth and layoutHeight has finished being updated ONLY by a new document being loaded. 

Went with this option.

j-huff:
Well nevermind I guess I broke it again.

j-huff:
okay I'm pretty sure it's actually fixed now 😅

courtneypattison:
I made a PR #158 for moving the origin to the top left!

j-huff:
> I made a PR #158 for moving the origin to the top left!

I didn't realize this was desirable. I purposefully moved the origin to center.

j-huff:
I see. Makes sense to me.

Issue #141, opened 2024-03-24
Change origin of pdf to left side when more than one page


courtneypattison:
I'm just remembering why I didn't recenter the pdf as a default. It makes stitching more difficult because you can't see the first few pages. Maybe we should move the pdf back to being left aligned when there is more than one page.

PR #158, opened 2024-03-28
Refactor geometry to top left origin

courtneypattison:
As discussed in #141


j-huff:
looks good 👍

courtneypattison:
> looks good 👍
> 
> Edit: now that we've tried both ways to their fullest extent I suppose that we know for sure that this way is better lol

Awesome, thank you for looking it over!!! I really want to get the beta branch out there. I think I'm going to let it bake tonight and merge into main tomorrow.

PR #164, opened 2024-03-29
Changed perspective matrix to be in units of measure. 

j-huff:
In my opinion, the perspective matrix (and therefore localTransform) should be in units of measure rather than PDF pt units. This simplifies a lot of operations as you don't have to take into account the ptDensity.

I also changed what was formerly layoutWidth/layoutHeight to layoutWidthPt/layoutHeightPt. This is the layout size in PDF units. Now, layoutWidth/layoutHeight is in units of measure, which also simplifies some operations (such as the recenter button).

### DIFF
--- a/app/[locale]/calibrate/page.tsx
+++ b/app/[locale]/calibrate/page.tsx
@@ -8,7 +8,11 @@ import CalibrationCanvas from "@/_components/calibration-canvas";
 import Draggable from "@/_components/draggable";
 import Header from "@/_components/header";
 import PDFViewer from "@/_components/pdf-viewer";
-import { getPerspectiveTransformFromPoints, toMatrix3d } from "@/_lib/geometry";
+import {
+  getPerspectiveTransformFromPoints,
+  toMatrix3d,
+  scaleTranslation
+} from "@/_lib/geometry";
 import isValidPDF from "@/_lib/is-valid-pdf";
 import { Point } from "@/_lib/point";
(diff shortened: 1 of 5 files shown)


courtneypattison:
I disagree on this one! To me, this is adding complexity rather than simplifying things. As a simple measure, this PR is doubling lines of code rather than removing any. I like that you're trying to clean up the code though!

j-huff:
> As a simple measure, this PR is doubling lines of code rather than removing any

Would you like me to do it in less lines? 13 of the "added lines" are either comments, empty, reformatting, or the lines needed for the added feature of resetting the transform on unitOfMeasure change. I could also get rid of the idea of keeping layoutWidthPt and layoutWidth as separate items and make the code less readable , but save a few lines.

If we want to actually consider code complexity rather than something arbitrary like number of lines, you will see that this is change makes it so you only need to deal with the conversion from PDF units to real units in essentially one place, rather than having to do every single place you want to deal in real units. Right now, that doesn't result in a *huge* amount of complexity savings, but it makes adding features such as the delta position popup slightly less annoying.

Consider the change to the MeasureCanvas component as an example:

This
```
   let d = Math.sqrt(sqrdist(line[0], line[1])) / CSS_PIXELS_PER_INCH;
    if (unitOfMeasure == CM) {
      d *= 2.54;
    }
 ``` 
 Becomes this:
 ```   
    let d = Math.sqrt(sqrdist(line[0], line[1]))
 ```
 
 Which one is more complicated?

Issue #126, opened 2024-03-21
White line around calibration grid is confusing


courtneypattison:
The white line around the calibration grid is confusing people understandably because it can't be seen with the current setup. Should we change the light mode calibration grid to be black lines on white? Or should we Make the border of the calibration grid purple? Or something else?

PR #169, opened 2024-04-01
Calibration canvas refactor

courtneypattison:
Cleaned up the calibration canvas


courtneypattison:
There's a lot still to be done, but I'd love to hear opinions on this one before I go much further. 

TODO:
- Press and hold arrow key will continue to move corners
- Clean up code to remove duplication of logic from hooks and share progressive increase
- Make it draggable? Maybe with a move icon to show it can be dragged. Not sure about this one. It feels really nice where it is on a tablet.
- Maybe add the toggling button to the right side like the layers menu?
- Save visibility of pad in localStorage
- Maybe name change? Not sure about movement pad.

hluedeke:
I just took a peek at the calibration canvas. Wow, a lot has changed! I see the grid is now white on black, the corners no longer have color - not sure when any of that happened 😅

(The colors on the corners were intended to help people figure out which corners matched on their projected space, in the case that their screen was flipped from their computer screen. Not sure if it actually helped or not, but that was the original purpose)

Here are my comments on the above:
- Can we detect a touchscreen? I remember this being dubious from my research, but I think we will want to only show the buttons on touchscreens (I'm assuming you are leaving them on desktop for now for testing purposes, but just wanted to make this note)
- I think that holding down the button should behave exactly like holding down the mouse. Having to tap a million times to line up the corners is going to be annoying if the touchscreen user has to go long distances. Maybe this is what your third point is referring to?
- As far as dragging, it's going to be tough to communicate that one, especially since all the functions are already touch handlers. My opinion: let's leave it be for now and see if anyone complains!
- I think the toggle button is potentially a better option, but I'll say same thing: let's start without it and see if anyone complains!
- Movement pad makes sense to me, other alternatives might be: touchscreen movement, movement buttons. I don't really know. I am terrible at naming things!

courtneypattison:
> I just took a peek at the calibration canvas. Wow, a lot has changed! I see the grid is now white on black, the corners no longer have color - not sure when any of that happened 😅

The grid inversion on light mode was because of #126. I had a few people complain. Fix happened on #138 :)

> 
> (The colors on the corners were intended to help people figure out which corners matched on their projected space, in the case that their screen was flipped from their computer screen. Not sure if it actually helped or not, but that was the original purpose)

I took these out because it's not easy to see the difference in colours on my projector. I was originally planning on replacing them with some symbol to mark each corner as unique but then realized that the dimension labels are probably enough to orient someone with a weird setup. The dimension labels are nice because text has an orientation so you know where the corner is. What do you think?

> 
> Here are my comments on the above:
> 
> * Can we detect a touchscreen? I remember this being dubious from my research, but I think we will want to only show the buttons on touchscreens (I'm assuming you are leaving them on desktop for now for testing purposes, but just wanted to make this note)

At first I was going to try to detect this but the more I think about it the more I like leaving it there for all cases and someone can simply toggle it off once if they don't want to use it. Two main reasons: someone using their mouse on their cutting mat without having to reference their device and someone who doesn't know about the arrow key movement. What do you think?

> * I think that holding down the button should behave exactly like holding down the mouse. Having to tap a million times to line up the corners is going to be annoying if the touchscreen user has to go long distances. Maybe this is what your second point is referring to? EDIT: Wow, I can't read. That is exactly what your first point says 😅 in short, yes I agree!

Lol, great! It's going to take me a minute to refactor the two and set up the intervals. React really boggles my mind on things like this.

> * As far as dragging, it's going to be tough to communicate that one, especially since all the functions are already touch handlers. My opinion: let's leave it be for now and see if anyone complains!

I'm happy with that!

> * I think the toggle button is potentially a better option, but I'll say same thing: let's start without it and see if anyone complains!

There is a toggle button in the menu bar. Do you think I should remove that for now?

> * Movement pad makes sense to me, other alternatives might be: touchscreen movement, movement buttons. I don't really know. I am terrible at naming things!

How about move tool?

> 
> === EDIT
> 
> Another note:
> 
> Maybe this icon for the center button, possibly rotated?
> 
> https://fonts.google.com/icons?selected=Material%20Symbols%20Outlined%3Arefresh%3AFILL%400%3Bwght%40400%3BGRAD%400%3Bopsz%4024
> 
> Or maybe this one?
> 
> https://fonts.google.com/icons?selected=Material%20Symbols%20Outlined%3Arefresh%3AFILL%400%3Bwght%40400%3BGRAD%400%3Bopsz%4024
> 
> I thought the center button was going to center the canvas 😅 just an idea, might not be that important.

I'm seeing the same refresh icon for both those links but I like that!

hluedeke:
> I took these out because it's not easy to see the difference in colours on my projector. I was originally planning on replacing them with some symbol to mark each corner as unique but then realized that the dimension labels are probably enough to orient someone with a weird setup. The dimension labels are nice because text has an orientation so you know where the corner is. What do you think?

Ooh, that's a good point. I hadn't considered the new text - that will probably fix it!

> At first I was going to try to detect this but the more I think about it the more I like leaving it there for all cases and someone can simply toggle it off once if they don't want to use it.

Maybe default on for touchscreen and default off for desktop? I was really annoyed by it being there when I first opened up the page. It felt like unnecessary screen usage (but admittedly, I'm on a desktop with a mouse right now).

> Two main reasons: someone using their mouse on their cutting mat without having to reference their device

I don't understand the use case here. Are you saying they would rather use the mouse to click on the buttons than the mouse to click on the corners? I guess MAYBE I could see that. I am still a little skeptical that we should use up screen space (by default) for it.

> and someone who doesn't know about the arrow key movement. What do you think?

I'm also a little skeptical on this. I don't think we should justify using screen space to lower the learning curve even further than it already is. I think the vast majority of people "get it" without too much trouble. My personal opinion!

> There is a toggle button in the menu bar. Do you think I should remove that for now?

If there's already a button, then say no more! Let's just leave it in there!

I did notice on some of Sasha's help posts that her screen is cutting off several of the buttons. That also can be fixed by menu management, though. I wish I had more time to work on this right now! I would love to tackle menu management.

This was the alternate icon. Sorry about that!

https://fonts.google.com/icons?selected=Material%20Symbols%20Outlined%3Acycle%3AFILL%400%3Bwght%40400%3BGRAD%400%3Bopsz%4024

courtneypattison:
I think you're right! The majority of people have keyboard setups where the nicest way to calibrate is with the arrow keys and tab. I'll see if I can reliably determine if people are using touch screens and conditionally show. If not, I'll hide it by default. I may move the toggling button in with the movement pad so it's more obvious how to toggle it.

hluedeke:
> Lol, great! It's going to take me a minute to refactor the two and set up the intervals. React really boggles my mind on things like this.

I also wanted to throw in a bit of a tidbit here - with the existing hooks, you should be able to abstract out the event type and pass it in as an argument ("onmousedown", "ontouchtap" <= or whatever the touch event is) then reuse the logic completely! You shouldn't have to do too much here (at least I believe this is the case!)

courtneypattison:
I think it's a bit more complicated because onClick and onPointerDown events don't repeatedly fire on hold like onKeyDown so I believe I need to create my own interval timer to fire until onPointerUp. Ideally with the same speed as the onKeyDown repeat to have the same feel as the arrow key movement. Please correct me if I'm wrong!

hluedeke:
> I think it's a bit more complicated because onClick and onPointerDown events don't repeatedly fire on hold like onKeyDown so I believe I need to create my own interval timer to fire until onPointerUp. Ideally with the same speed as the onKeyDown repeat to have the same feel as the arrow key movement. Please correct me if I'm wrong!

Oh darn it. That's right. I ran into the same thing with the "hide after 3 seconds" logic we were trying to implement. I started experiencing performance lag, unfortunately, even using recommended React methods for debouncing. I should get my branch up there so you can take a look at it - the code will be similar. But, as I said, it currently suffers from drawing/performance issues (which jhuff also ran into).

In fact, I'm a little worried that we are going to start running into performance bottlenecks in general. The issue where on hover text stops working after awhile smells strongly of performance/memory limitations.

hluedeke:
...ok, now I'm thinking about it a bit...

Wouldn't you use onmousedown/onmouseup? It also appears there is onTouchStart and onTouchEnd. I think those might solve your debouncing problem...

courtneypattison:
> ...ok, now I'm thinking about it a bit...
> 
> Wouldn't you use onmousedown/onmouseup? It also appears there is onTouchStart and onTouchEnd. I think those might solve your debouncing problem...

That's my plan! onPointerDown -> onPointerUp. I started working on it but ran into the problem described [here](https://itnext.io/how-to-work-with-intervals-in-react-hooks-f29892d650f2). I'm planning on using a reducer for the points to solve the problem. We'll see how that goes!

I got a React book yesterday to learn a bit more about fixing some of these performance issues and using better design patterns. Hopefully it helps!

courtneypattison:
I believe I added all the features discussed. I used a slightly different pixel list for the progressive increase on the movement because it felt too fast when I tested it out on a tablet. Probably because it fires less frequently than with the keyDown. Please let me know if you think I should change those numbers! If you think it looks ok, I'm happy enough with merging this one into beta.

courtneypattison:
I accidentally added the fix for #170 on this branch. Since I'd like to get that one out there and am too lazy to switch it to beta, I'm just going to merge this PR into beta. If you notice any issues with this one, I'll just fix them directly on beta! Sorry for the clumsiness!

PR #205, opened 2024-04-11
Sliding window erode for Safari.

courtneypattison:
Use a sliding minimum window and separate horizontal and vertical passes to improve performance for Safari.


courtneypattison:
Closing as #235 is way better lol

PR #208, opened 2024-04-12
Fullscreen and window movement checks

courtneypattison:
Will display pop ups when the window has moved or changed size when projecting compared to the window size/location when calibrated.


courtneypattison:
@hluedeke @sashasewist If you two have the time, could you please look at this preview? I've added checks whenever there is a screen size or movement when calibrating. I'm really not sure about the wording of the popups and would love input on those. I think these checks should be sufficient for making sure people have accurate scale. Thoughts?

Issue #128, opened 2024-03-22
Maximum update depth exceeded warning (cursor timeout)


j-huff:
on the beta branch, I'm getting a warning when dragging the the pdf:

Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.
eval	@	draggable.tsx:83
eval	@	draggable.tsx:96

It has something to do the resetIdle and handleOnMouseMove functions but I couldn't figure out how to fix it. Tried changing them to useCallback functions but that didn't seem to work.

courtneypattison:
I can't reproduce this one.

j-huff:
Yeah it seems to only be happening on my home Linux desktop, which is pretty slow. Maybe not a real issue?

courtneypattison:
Maybe close for now and reopen if it causes problems in the future?

hluedeke:
I was able to reproduce this - will look into it

PR #221, opened 2024-04-19
Add transform context to fix menu flickering

### LINKED ISSUE #220: Menu flickering when dragging large PDF
I think this is happening because the whole page is re-rendered when the PDF is moved. I want to do a big refactor of the whole site based on some new info I have about React design patterns/performance. This issue seems like the place to start.

courtneypattison:
Fix #220! Woot woot!

TODO
- [x] Layer menu not opening in Firefox and Safari
- [x] Safari unresponsive
- [x] Firefox line weight broken


courtneypattison:
I ended up removing the shift to axis lock when dragging PDF. I think it makes more sense just to use the arrow keys to move along a horizontal or vertical axis. It was slowing the dragging down, especially in Safari for large PDFs and was a desktop only solution.

courtneypattison:
I believe this PR also fixes #128

PR #224, opened 2024-04-19
Beta

courtneypattison:


courtneypattison:
@sashasewist @hluedeke @roel-v I think the beta branch is nearly ready to be merged into main. If any of you have time, I would love to get some feedback before merging it in. I'm going to message translators as well and hopefully have it ready in the next day or two. The [CHANGELOG](https://github.com/Pattern-Projector/pattern-projector/blob/main/CHANGELOG.md) should have a fairly good overview of what has changed, but I definitely could've missed a few things.

What do you think about renaming the "mark, measure, and transform" tool  "line tool"? Too vague?

courtneypattison:
iPad bugs
- [ ] Frozen when try to pop back into PDF after opening when using browser instead of app
- [ ] Timer on stitch stepper not being canceled
- [ ] Cannot drag end of line tool
- [ ] Text cursor highlights when pressing stepper 
- [ ] Too easy to trigger selecting text in stepper when trying to press +/-

PR #228, opened 2024-04-20
Add scale tool

courtneypattison:
Add scale input. I'm not happy with the input field, but I'm not sure how to change it to be better. Anything with a input box is no good for iPad because it pops you out of full screen whenever the input box is focused to display the keyboard. This would be incredibly frustrating to use on iPad since people seem to want to adjust scale on the fly looking at how it will change in real time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Scale tool
 - Full screen button to projection page
 - Cursor changes from grab to grabbing hand when dragging
 - Selected corner on calibration grid is colored
--- a/app/[locale]/calibrate/page.tsx
+++ b/app/[locale]/calibrate/page.tsx
@@ -88,7 +88,7 @@ export default function Page() {
     horizontal: "0",
     vertical: "0",
   });
-
+  const [patternScale, setPatternScale] = useState<number>(1);
   const [menuStates, setMenuStates] = useState<MenuStates>(
     getDefaultMenuStates(),
   );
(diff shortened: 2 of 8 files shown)


sashasewist:
Is a slider possible? Or even both? A slider to test out what scale will work best by projecting it in real time, and an input box if you know without testing what scale you want.

sashasewist:
A slider would need more room, and sort of related, I don't think scale adjustment should be in main toolbar. we would never want a misclick to change scale.
so how about where the ^ button is have two buttons:
^ (hide tools)
+ (more tools
then when + is pressed a second row of tools shows up, which would include scale, and others that are less commonly used, maybe measuring? also annotating tools when you add those?

courtneypattison:
> A slider would need more room, and sort of related, I don't think scale adjustment should be in main toolbar. we would never want a misclick to change scale. so how about where the ^ button is have two buttons: '^ (hide tools) '+ (more tools '(ignore apostrophes, just added them because the + was activating the markdown lol) then when + is pressed a second row of tools shows up, which would include scale, and others that are less commonly used, maybe measuring? also annotating tools when you add those?

Yes to all these ideas! I'd also like to have some sort of warning or style change when the scale is changed so people know it's dangerous. Some ideas for the warning: popup that explains scale should only be changed purposefully not to calibrate but only show this once and have them click a button to ok it (save that they've seen this message in localStorage like calibration), change the colour of the scale number/text to red, show a warning that cannot be clicked away from but will disappear like the mouse after a few seconds of inactivity (is only shown when the mouse is moving or tapping happens). Not really sure about any of these, just want some sort of caution.

courtneypattison:
I added a more tools overflow menu that has the scale option. I'm thinking we should wait for the steppers being added from #233 and use those same ones for the scaling. That'd keep the UI fairly consistent. Still need to decide on a warning and/or UI change when the scale changes though.

Issue #200, opened 2024-04-09
Add an annotation/tag for shifting pattern


courtneypattison:
Someone from the group wants something to tag a spot then shift the patter/fabric when the piece is larger than the projection area. I made the measure tool keeping this sort of functionality in mind. I was thinking maybe dual purpose it by having the measurement move with the pattern. Is that too undiscoverable? Separate button for marking a spot? Something else?

courtneypattison:
Doodle overlay? A few unique shapes scattered around the viewport that move with the pattern, making landmarks to go by.

sashasewist:
> Someone from the group wants something to tag a spot then shift the patter/fabric when the piece is larger than the projection area. I made the measure tool keeping this sort of functionality in mind. I was thinking maybe dual purpose it by having the measurement move with the pattern. Is that too undiscoverable? Separate button for marking a spot? Something else?

You have mae a measure tool? I will check this out, have not seen it. Making that tool dual purpose sounds interesting. I will report back later after testing.

sashasewist:
I tried your measure tool, very nice!! Would be nice to have it be a 3 stage toggle:
- click ON, and mesure tool is active and shows measurements 
- click again and measurements are hidden but measured lins stay
- click again and both lines and measurements are hidden/OFF
- click ON and they come back.
It would be nice if the pattern could be scrolled when in measuring mode (and yes marks stay in same location on off). not just for when needing landmarks for shifting fabric, but also when measuring multiple parts of pattern pieces we need to scroll.
I found this icon that has ruler and pencil, but scale is likely not right for visibility in PP toolbar:
https://thenounproject.com/icon/ruler-5088326/

courtneypattison:
I like that idea! I’ve been thinking something of something similar. I at least want the lines to stay with the pattern when moving. I think having the different modes would be a great solution.

courtneypattison:
I'm getting excited about this one lol. We use material icons in PP. I found one similar to the one you found https://fonts.google.com/icons?selected=Material+Symbols+Outlined:design_services:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=pencil

I'm thinking draw whatever lines you want and they stay on the pdf and move with it even when the measure tool isn't activated. Click on a line and it'll show context for that line like length, angle, rotate to horizontal button, and delete button to get rid of the line.

sashasewist:
Sounds perfect! 👍🏻

Issue #239, opened 2024-05-01
Maybe show disabled layer menu icon when no layers


courtneypattison:
I’m thinking maybe we show the layers menu icon greyed out when there are no layers. Maybe also change the tooltip for it to say no layers in this pattern? I’ve been confused when I bought a pattern with no layers and didn’t see the menu. Seems to keep coming up.

sashasewist:
That sounds like a good idea.

PR #242, opened 2024-05-03
Add mark and measure tool

### LINKED ISSUE #178: Escape exits full screen instead of exiting measure
I think it'd be best just to remove the escape from measure and not overload the exit full screen. m can be used to toggle measure off anyway.

courtneypattison:
As per #200. @sashasewist when you get a chance, please check this preview out! It allows drawing multiple measurement lines that move with the PDF. x to delete last line and arrow button to rotate PDF to horizontal.

Fix #178


sashasewist:
Love it!
- I tested in Firefox, and I get a full screen message flashing for half second when I use it. I didn't calibrate or project in full screen.
- I love that shift makes the line straight! Is is possible to have esc key turn off drawing tool and go back to regular mode? It gets annoying when it keeps drawing when I want to move pattern or select a line instead.
- the Click on a line and it will show context is not quite working. When tool is active it only wants to draw new line, not select existing line.

courtneypattison:
> * I tested in Firefox, and I get a full screen message flashing for half second when I use it. I didn't calibrate or project in full screen.

Good catch! I’ll check that out. 

> * I love that shift makes the line straight! Is is possible to have esc key turn off drawing tool and go back to regular mode? It gets annoying when it keeps drawing when I want to move pattern or select a line instead.

I originally had esc for this exact case. Except when I was testing using full screen I realized it would exit full screen instead. I’m thinking maybe exit the measure tool mode after each line is drawn. I’ve also been annoyed by the measuring continuing! I’ll try that out and see if it feels ok. 

> * the Click on a line and it will show context is not quite working. When tool is active it only wants to draw new line, not select existing line.

lol, that’s because I haven’t implemented that feature yet. I got too excited to share that didn’t finish everything! I’ll let you know when I get these issues fixed up!

courtneypattison:
@sashasewist I'm having trouble reproducing the warning flashing on Firefox. Does it happen every time? Are you able to show a screencast of it? Also, I think this feature is about ready to go. Could you please take a look when you get a chance to see if it needs anything else? Thank you!

sashasewist:
Tested again, works great! One slight issue is that when stitching function is active (not yet closed), other tools work, but the draw measure tool does not. And it makes me think... It could actually be a super useful tool to  use when stitching, in that mode, the measurement could also be shown in pixels, so that people could use the tool to know what to enter in the HV boxes.

sashasewist:
The rotate function integrated is so helpful, you are brilliant to see the need for that, to bring A0 pieces on grain. And if we draw a 45 degree line, will be perfect for cutting pieces on bias! I keep trying to think of other wording than "rotate to horizontal", as I don't think people will understand that. I think they will think line will rotate, not the pattern. But it's OK< we will encourage people to try it out to understand how useful it is.

courtneypattison:
@sashasewist thank you for checking this out! You've got me thinking about changing it more, lol. I love the stitch menu edge measuring idea. Not sure how to get that interaction nice yet. I'll ping you again with the next iteration!

sashasewist:
Another thought , the tool tip says "Start measuring" but it's so much more. Could it say "Extra tools" or "Advanced tools". Or if you have room - "Extra tools: Mark, Measure, and Rotate grain".

courtneypattison:
Lol, let's call it whips and whistles! I agree that It should be more descriptive. I like "Mark, measure, and rotate grain" I think that sums it up pretty well for now! Happy Mother's Day!

von-matti:
I just tested on Microsoft Edge. Worked very well for me. Just that measuring isn't in cm and the description is english while all the other tools are in german :)

courtneypattison:
> I just tested on Microsoft Edge. Worked very well for me. Just that measuring isn't in cm and the description is english while all the other tools are in german :)

@von-matti thank you so much! Did you calibrate in cm? The tool should measure using whatever measurement you use to calibrate. It won’t be in translated to German yet. Once we add the feature to the beta site, then it’s ready to be translated and I reach out to a few people to have it translated. We use weblate to translate if you’re interested in contributing and bilingual! https://hosted.weblate.org/projects/pattern-projector/pattern-projector/

courtneypattison:
@sashasewist I think this one may finally be ready to go. I haven't added in the measuring for stitch edges. That needs more thought, but I opened #251 for it. I had a bit of fun and added flipping along the line and moving based on the line length (for alterations). I also added moving the ends of the line to be able measure things larger than the screen as per #249 @von-matti's suggestion.

A couple other small things: fixed #250 and added #239.

Issue #233, opened 2024-04-23
Add steppers to the stitch menu inputs


courtneypattison:
Steppers only show up on computer setups. I'd like to make something that looks the same and shows up on all devices.
e.g. [-]1[+]

hluedeke:
I can take a look at this one!

PR #246, opened 2024-05-11
First commit for #233. Adds a stepper component to stitch menu for al…

hluedeke:
…l numeric inputs. Will require testing on a tablet to ensure the steppers are easy to click/use

### DIFF
--- a/app/_components/stepper-input.tsx
+++ b/app/_components/stepper-input.tsx
@@ -50,7 +50,7 @@ export default function StepperInput({
         <button
           type="button"
           name={name}
-          className="absolute top-0 left-0 h-full flex items-center"
+          className="absolute top-0 left-1 h-full flex items-center"
           onClick={() => onStep(-1)}
         >
           <StepDownIcon ariaLabel="test" />
@@ -58,7 +58,7 @@ export default function StepperInput({
         <input
           min={min}
           type={type ? type : "text"}
-          className={`${inputClassName} text-center p-2.5 text-gray-900 text-sm focus:outline-none dark:bg-black dark:placeholder-gray-400 dark:text-white`}
+          className={`${inputClassName} bg-gray-50 text-center p-2.5 border border-gray-300 dark:border-gray-600 text-gray-900 text-sm rounded-lg focus:ring-4 focus:outline-none dark:bg-black dark:placeholder-gray-400 dark:text-white dark:focus:ring-blue-800 dark:focus:border-blue-800`}
           id={id}
           data-test-id={inputTestId}
           name={name}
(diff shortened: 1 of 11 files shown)


sashasewist:
Nice! Is it possible to keep a box around the H and V values? I think a box would make it more clear that we can still type numbers.

sashasewist:
And one more suggestion - would "Pages to assemble" instead of just "Pages" help people understand what this screen is for, and to make them realize they should change pages to skip the tutorial pages?

hluedeke:
> And one more suggestion - would "Pages to assemble" instead of just "Pages" help people understand what this screen is for, and to make them realize they should change pages to skip the tutorial pages? (or "Pages to stitch"?)

How about "Page Range"? Isn't that roughly what Adobe says?

courtneypattison:
Ooo, thank you!

> Nice! Is it possible to keep a box around the H and V values? I think a box would make it more clear that we can still type numbers.

I agree it's not as obvious that you can type in numbers. Another idea is maybe show the cursor when an increment or decrement button is hit for that input?

> And one more suggestion - would "Pages to assemble" instead of just "Pages" help people understand what this screen is for, and to make them realize they should change pages to skip the tutorial pages? (or "Pages to stitch"?)
> How about "Page Range"? Isn't that roughly what Adobe says?

I think I see where you're both coming from. It's not obvious what the stitch menu does so changing that label would make it more apparent. What about changing it to "Stitch Pages" and moving that to the first spot in the stitch menu (swap with columns)? It would act kind of like a menu header/hint and be more informative.

In the long term, I'd love to get rid of that input field altogether, move layout/stitching into its own page, and make it more touch friendly/graphical. I have quite a few ideas for this one!

Also, I think the increment and decrement buttons need to repeat when held. That's mainly how I use the current browser based ones. @hluedeke you can see how I did that for the move tool. I had to change the state to a reducer and set a timer. If you don't have time to add this, just let me know and I can do it :)

courtneypattison:
@hluedeke have you had a chance to work on this PR? I'd like to try to release a new version sometime in the next few days since I really want to get #247 out there! If you don't have time, I can look at it :) Thank you for getting the ball rolling on this one!

hluedeke:
> @hluedeke have you had a chance to work on this PR? I'd like to try to release a new version sometime in the next few days since I really want to get #247 out there! If you don't have time, I can look at it :) Thank you for getting the ball rolling on this one!

Actually, I had worked on a bit more, I just hadn't committed it yet. To be honest, I have been so busy that I'm not sure where this code ended up. But I added at least some of the suggestions from @sashasewist. I went ahead and updated the PR so you can see what I ended up with!

hluedeke:
I remember what I was going to work on - at the moment, if you hold down the + or - buttons, nothing happens. Ideally, if you hold down those buttons, then the increase/decrease would still happen but at a faster rate. At the moment you have to click the +/- every time you want to increase or decrease the number. I just never got to that part - sorry!

PR #279, opened 2024-06-09
Always show stitch menu with export PDF option

courtneypattison:
## Pull Request Checklist

## Description

- Changed save to export
- Moved stitch input labels above boxes to make more room horizontally
- Added tooltip to export button
- Always show stitch menu button regardless of page count (disabled until PDF is opened)
- Removed "projecting/calibrating" label from menu to make more room
- It's possible to press all buttons in the menu, stitch, and layers menu on an iPhone SE when in landscape. Bad UX but at least it's possible
- I decided not to add an experimental tooltip since it's working really well! I think people get that anything new we put out will have some issues.

Making the stitch menu always available and having the export button in there is the best solution I could come up with for always having the button available while working within space constraints. I think it also makes sense to keep that functionality together, like a mini pdfstitcher in PP. That way if/when we pull it out it's all together.

Please let me know what you think @sashasewist and @cfcurtis. Sorry for taking so long to get this out! Also, @cfcurtis would you be ok with me adding you to the project?

## Related Issues

#259

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Download PDF button that saves the PDF with selected layers and stitched pages
+
+### Changed
+
+- Always show stitch menu button regardless of the number of pages in the PDF
+
 ## [1.0.2] - 2024-05-25
 
 ### Changed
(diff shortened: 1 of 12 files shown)


cfcurtis:
This looks awesome! I just tested on my phone and all the UI elements were visible and functional, and I was able to stitch and export a PDF. Saving layer visibility also worked, we'll just need to tell people to click on the stitch menu first if there's only one page.

I'd be happy to be added to the project, and I can continue to support things related to this. Thanks!

courtneypattison:
Awesome, thank you for testing @cfcurtis! I'm going to merge this into beta :)

Issue #272, opened 2024-05-27
Save layers and position in localStorage


courtneypattison:
So people don't have to input layers and move for same pattern

courtneypattison:
I've left the code, but temporarily removed saving the transform because I'm worried about there being no reset for any scale issues I may have introduced with the magnify and zoom out. I'll add it back in when I'm more confident about those modifications.

PR #281, opened 2024-06-14
Store visible layers in local storage per file

Nifdee:
## Pull Request Checklist

## Description

Whenever a document is opened for the first time, or its visible layers change, the set of visible layers now gets stored to local storage. When the user re-opens the same PDF again, that setting is used to initialize which layers will be visible initially.

## Related Issues

First part of https://github.com/Pattern-Projector/pattern-projector/issues/272


app/_components/pdf-custom-renderer.tsx
```diff
@@ -81,32 +80,9 @@ export default function CustomRenderer() {
     }
     async function optionalContentConfigPromise(pdf: PDFDocumentProxy) {
```
Nifdee, line 81: This is now only used for configuring how the pdf should be rendered, i.e. telling pdfjs which layers to show. Initializing which layers the document has now happens in PdfViewer.

app/_components/pdf-viewer.tsx
```diff
@@ -59,14 +59,14 @@ export default function PdfViewer({
...
     setPageCount(numPages);
-    setLayers(new Map());
     setPageSize({ action: "clear" });
     if (stitchSettings.pageRange.endsWith("-") && numPages > 0) {
       dispatchStitchSettings({
         type: "set-page-range",
         pageRange: stitchSettings.pageRange + numPages,
       });
     }
+    getLayersFromPdf(docProxy).then((l) => setLayers(l));
```
Nifdee, line 69: It felt nicer to do this here, rather than first setting an empty map and then setting the layers during rendering.

app/_lib/layers.ts
```diff
@@ -0,0 +1,28 @@
+import type { PDFDocumentProxy } from "pdfjs-dist";
+import { Layer } from "./interfaces/layer";
+
+export type Layers = { [key: string]: Layer };
```
Nifdee, line 4: This is easier to work with in the reducer since the reducer needs to be pure.

courtneypattison:
Wow, such an awesome PR! Thank you so much! I really appreciate you cleaning up and rearchitecting along the way.

One tiny change is to base off the beta branch. That's where we put everything for staging before a release.

I don't want to get too greedy but would you like to be added to the project? That way you could assign issues to yourself if you'd like to work on anything else. No pressure! I'm just really happy with this PR :)

courtneypattison:
I opened #282 to fix a few build errors I got from switching the base of this PR to beta. I'm going to close this one and merge the other into beta :)

Nifdee:
Yay, thank you!

> One tiny change is to base off the beta branch. That's where we put everything for staging before a release.

Good to know - I'll update the instructions to include that and base off beta going forward :)

> would you like to be added to the project?

I'd love to! You can't imagine how excited I was when I found this project :D

courtneypattison:
> > One tiny change is to base off the beta branch. That's where we put everything for staging before a release.
> 
> Good to know - I'll update the instructions to include that and base off beta going forward :)

Thank you so much! I should've done that a long time ago!

> 
> > would you like to be added to the project?
> 
> I'd love to! You can't imagine how excited I was when I found this project :D

 Woot woot! I'll send and invite :)

PR #283, opened 2024-06-14
Configure prettier to always use trailing comma

Nifdee:
## Pull Request Checklist

## Description

Without this change, the automatic formatting on save would drop some commas that would get re-added by running `yarn build`. It also wasn't obvious which prettier extension I needed to install, so I added this one as a recommended extension for future people coming across this.


courtneypattison:
Thank you so much!

PR #285, opened 2024-06-15
Fix bug where visible layers would reset

Nifdee:
## Pull Request Checklist

## Description

While working on something else, I noticed the new "save visible layers" functionality had a bug in the following scenario:
* Open a PDF, customize which layers to show / hide.
* Open a different PDF.
* Open the first PDF again, but close it without toggling any layers.
* Reload the page, or move between PDFs again - now all layers will be visible again

This PR fixes that bug.


courtneypattison:
Ooo, good catch! Thank you!

PR #287, opened 2024-06-17
Center single page PDF on load

josephineb:
## Pull Request Checklist

## Description

When a single page PDF is uploaded, it is automatically centered in the PDF viewer.

## Related Issues

#278

### DIFF
--- a/app/_components/pdf-viewer.tsx
+++ b/app/_components/pdf-viewer.tsx
@@ -15,10 +15,12 @@ import CustomRenderer from "@/_components/pdf-custom-renderer";
 import { getPageNumbers } from "@/_lib/get-page-numbers";
 import { PDF_TO_CSS_UNITS } from "@/_lib/pixels-per-inch";
 import { RenderContext } from "@/_hooks/use-render-context";
+import { useTransformerContext } from "@/_hooks/use-transform-context";
 import { StitchSettings } from "@/_lib/interfaces/stitch-settings";
 import { StitchSettingsAction } from "@/_reducers/stitchSettingsReducer";
 import { getLayersFromPdf, Layers } from "@/_lib/layers";
 import { LoadStatusEnum } from "@/_lib/load-status-enum";
+import { Point } from "@/_lib/point";
 
 pdfjs.GlobalWorkerOptions.workerSrc = new URL(
   "pdfjs-dist/build/pdf.worker.min.js",
@@ -39,6 +41,7 @@ export default function PdfViewer({
   filter,
   setPdfLoadStatus,
   setLineThicknessStatus,
+  gridCenter,
 }: {
   file: any;
   layers: Layers;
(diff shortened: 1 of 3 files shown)


courtneypattison:
Looks great! Thank you @josephineb!

Since we only recenter when there's one page, the layoutWidth and layoutHeight are just the width and height of the page. Therefore, we can use those sizes on page load success. Something like this:
```
function onPageLoadSuccess(pdfProxy: PDFPageProxy) {
    const scale = (pdfProxy.userUnit || 1) * PDF_TO_CSS_UNITS;
    const width = pdfProxy.view[2] * scale;
    const height = pdfProxy.view[3] * scale;
    setPageSize({
      action: "setPage",
      pageNumber: pdfProxy.pageNumber,
      width,
      height,
    });
    if (pageCount === 1) {
      transformer.recenter(gridCenter, width, height);
    }
  }
```

That way we don't need any additional state! What do you think?

courtneypattison:
It's a bit of a special case. We may need to add the state back in the future but for now I think it's a bit simpler. Thank you so much for the quick fix! Would you be able to remove `transformer` and `grid center` from the `useEffect` as well please since those aren't being used anymore. Then I can merge this PR in :)

josephineb:
> It's a bit of a special case. We may need to add the state back in the future but for now I think it's a bit simpler. Thank you so much for the quick fix! Would you be able to remove `transformer` and `grid center` from the `useEffect` as well please since those aren't being used anymore. Then I can merge this PR in :)

Ah, I didn't clean that up. I think it's good to go now ;)

app/_components/pdf-viewer.tsx
```diff
@@ -72,12 +75,17 @@ export default function PdfViewer({
 
   function onPageLoadSuccess(pdfProxy: PDFPageProxy) {
     const scale = (pdfProxy.userUnit || 1) * PDF_TO_CSS_UNITS;
+    const width = pdfProxy.view[2] * scale;
+    const height = pdfProxy.view[3] * scale;
     setPageSize({
       action: "setPage",
       pageNumber: pdfProxy.pageNumber,
       width: pdfProxy.view[2] * scale,
       height: pdfProxy.view[3] * scale,
```
Nifdee, line 84: Super nit, but we can reuse the `width` and `height` values from above :)
```suggestion
      width,
      height,
```

josephineb: > Super nit, but we can reuse the `width` and `height` values from above :)

You're absolutely right! I have made the change. I apologize for the late reply; I was on holiday and unable to address this sooner. I also rebased the branch, resolving some conflicts. Please review it to ensure my changes are still correct :smiley:

courtneypattison:
Sorry for not merging this earlier! I read that nit and then left this PR and forgot. It must've been annoying fixing the merge conflicts. Thanks again! It's going to be really nice not having the recenter with large PDFs.

PR #290, opened 2024-06-21
Close overlay menu on click outside

Nifdee:
## Pull Request Checklist

## Description

I don't fully understand the cause, but clicks outside the overlay options menu would only trigger the `mousedown` event when the click was inside the header - not when it was on the main part of the page where the PDF gets rendered. The `pointerdown` event, however, _does_ get triggered everywhere so this PR switches to using that event everywhere.
Since there are three dropdown components that all run the same logic, I've extracted a hook for this.

## Related Issues

Fixes https://github.com/Pattern-Projector/pattern-projector/issues/262


courtneypattison:
You're on fire!!! Thank you! Looks excellent as usual.

PR #295, opened 2024-06-29
Disable pdf links

hluedeke:
## Pull Request Checklist

## Description

Disables PDF links by disabling rendering of the PDF annotation layer. This was (roughly) the recommendation from PDFjs itself (https://github.com/mozilla/pdf.js/issues/10664). All relevant pattern data *should* be in the content layer, however, if we are having issues with disabling the annotation layer, we can always try a custom css option (if we have a way to implement custom css)


courtneypattison:
Ooo, lovely to see your PR @hluedeke!

I had turned off the annotation layer when we were having size issues in the first week or two after the release. Then a couple of months ago someone was complained that they couldn't see their annotations in PP--they put many notes in their patterns--so I turned it back on.

Now I feel conflicted! What do you think?

hluedeke:
Hi! I know! Feels good to be back! 😄

If the annotation layer is important, than we can try the custom css route. I just need a way to define custom css in the project (aka create our own css classes). Do we have a place to do this?

courtneypattison:
> If the annotation layer is important, than we can try the custom css route. I just need a way to define custom css in the project (aka create our own css classes). Do we have a place to do this?

There's `app/[locale]/globals.css` that you can throw custom css classes in: [how to from tailwind!](https://tailwindcss.com/docs/adding-custom-styles#using-css-and-layer)

courtneypattison:
Perfect, thank you @hluedeke!

Issue #356, opened 2024-12-09
Exported PDF overstitched


courtneypattison:
Reported by Sheredith, who said a few people have reached out over this issue. The pattern has extra overlap when exported. @cfcurtis I'll try and take a look at this issue when I have a chance but figured I'd tag you in case you have any ideas. The first thing that comes to mind is maybe the overlap isn't getting converted to PDF pixels and just the CSS pixels are being used, i.e. 96 pixels per inch for CSS (what user inputs) should be converted to 72 pixels per inch for the PDF export. https://github.com/Pattern-Projector/pattern-projector/blob/9a72f843eb7375e2f00962b443b9bd764b641a87/app/_lib/pixels-per-inch.ts#L3 Here's the code for going from PDF to CSS pixels that the renderer uses.

Edit: bad theory. I just checked and the input is in PDF pixels per inch

Here's a pattern it's happening for: https://fabrics-store.com/sewing-patterns/noor-wrap-dress-pattern-type-pdf?srsltid=AfmBOooucbYhJuRO-OilO7CugJ_QIrHjioKb7O6dZWeapWrugNaMI2rR

cfcurtis:
Thanks for the pattern example! I happened to have time to check this right away, and I think I know what's happening.

The pattern pieces are a different size (7.5 x 10") from the instructions (8.5 x 11"). The very last page is a "thank you" page that is also 8.5x11. If you open it in PP and stitch only pages 5-55, you'll see that no trim/overlap is needed and everything behaves as expected.

Maybe pop up a warning if there are mixed dimensions in the pages selected for stitching?

courtneypattison:
@cfcurtis wow you're fast! Thank you! I think a warning is a great idea. I'll leave this issue open as a to-do for that so I don't forget!

PR #357, opened 2024-12-10
Stitch overlap

courtneypattison:
## Pull Request Checklist

- [x] I have added documentation for any new functionality from this pull request.

## Description

Change page size in PDFStitcher lib to be the max page width and height in the user specified page range to match how the [PDF viewer works](https://github.com/Pattern-Projector/pattern-projector/blob/9a72f843eb7375e2f00962b443b9bd764b641a87/app/_components/pdf-viewer.tsx#L222)

## Related Issues

#356
After sleeping on it, I think rather than a popup, this fix makes sense. It's not ideal, but at least it matches how the viewer behaves!

### DIFF
--- a/app/_lib/pdfstitcher.ts
+++ b/app/_lib/pdfstitcher.ts
@@ -28,16 +28,18 @@ function trimmedPageSize(
 ) {
   /**
    * Computes the size for each trimmed page.
-   * Assumes that all selected pages are the same!
+   * Chooses the largest page width and height from the user specified page range to match how the pdf viewer works.
    */
-  const firstRealPage = (pages.find((p) => p > 0) || 1) - 1;
-  const firstPage = inDoc.getPage(firstRealPage);
+  let width = 0;
+  let height = 0;
+  for (const page of pages) {
+    const p = inDoc.getPage(page - 1);
+    const pageSize = p.getTrimBox() || p.getMediaBox();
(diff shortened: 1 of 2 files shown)


app/_lib/pdfstitcher.ts
```diff
@@ -28,16 +28,18 @@ function trimmedPageSize(
...
    */
-  const firstRealPage = (pages.find((p) => p > 0) || 1) - 1;
-  const firstPage = inDoc.getPage(firstRealPage);
-
-  const pageSize = firstPage.getTrimBox() || firstPage.getMediaBox();
-  const width = pageSize.width - settings.edgeInsets.horizontal;
-  const height = pageSize.height - settings.edgeInsets.vertical;
+  let width = 0;
+  let height = 0;
+  for (const page of pages) {
```
cfcurtis, line 35: I would suggest filtering for pages > 0 in case someone adds a blank space, otherwise `getPage` throws an error.

cfcurtis:
Makes sense, consistency is good!

I just have one comment about checking the page number validity, otherwise seems like a simple fix.

app/_lib/pdfstitcher.ts
courtneypattison: Ooo, yes, silly me. Thank you!

courtneypattison:
@cfcurtis Thank you so much for reviewing this PR! I have fixed the bug you spotted and will merge into beta soon.

Issue #376, opened 2025-03-22
Change scaling to percentage


hluedeke:
I think it would be easier for people to understand scaling if it was in terms of percent (eg 100%) instead of just scale. All we need to do here is divide the input box by 100 when applying scale, and multiply the actual scale by 100 to populate the input box

PR #377, opened 2025-03-22
Scale to percentage

hluedeke:
## Pull Request Checklist

- [x] I have added documentation for any new functionality from this pull request.

## Description

This PR updates the scale menu to use percentages instead of decimal for scale. Percentage zoom is a much more familiar for those switching from Adobe.

- Modifies the location of the scale message (please give feedback).
- Modifies the "Open PDF" message slightly (adds styling)
- Moves some button styling into the global css file, as this is a better method for global styling than using functions

## Related Issues

#376 

## TODO

Review styling changes. You may disagree with them - would like feedback!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,12 +10,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Side menu for stitch, layers, and scale menus
-- Scale menu for changing the pattern scale
+- Scale menu for changing the pattern scale (modified to use percentage instead of decimal)
 - "Open With Pattern Projector" for PDF files on desktop, when installed with Chrome/Edge
 - Show a pop up when there's an error
 - Change page range in stitch menu with +/- (use + to throw beginning pages at the end e.g. for instruction pages)
 - Option to arrange stitched pages by column order
 - Support for SVG (with layer visibility toggling), PNG, and JPEG
+- Modified open pdf message slightly to add styling
 
 ### Changed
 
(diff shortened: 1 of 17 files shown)


hluedeke:
I am going to close the PR temporarily after talking with Sasha. Will reopen if I get to a good place!

hluedeke:
I ended up removing the conversion to percentage in the actual scaling form, but since I changed the styling for showing zoom, there is more room to show scale in both units (decimal scale and percent scale). This keeps the percentage zoom folks from asking questions but also (hopefully) prevents questions from those who are confusing on decimal.

courtneypattison:
Sorry for taking so long to get back to you @hluedeke ! The whole family has been battling a nasty virus the last few days so I haven't had any time for PP.

I mostly agree with your changes! The scale factor is definitely confusing for folks. As @sashasewist mentioned, the reasoning for not using percentage is that it's too likely that people coming from Adobe will be confused and input their percentage. I like the switch to .05 step. I'm almost thinking maybe even lower to a .01 step since the people who have requested scaling all seem to be doing tiny scale changes. What do you think? Also, with the percentage now being shown in the banner, the 1% step may be easier to understand.

Thank you for all the styling cleanup! It's much needed.

The open PDF message is much prettier and easier to understand, thank you!

The size and subtlety of the scale banner scares me a bit. I've read about people ignoring the calibration warning and still cutting even though its right in the middle of the screen. I think it makes sense to make it a banner and also maybe format the magnify/zoom out messages in the same way but I think it needs to be obnoxious and in people's faces. I'm not sure how to make a sufficiently eye catching banner but I'd like to see one because my current styling looks bonkers, lol. 

Thank you again! The code and front end always look cleaner after you send a PR!

Issue #412, opened 2025-12-16
All layers shown (for some PDFs?)


cfcurtis:
Reported on Facebook: all layers are displayed in Pattern Projector, even if the state is set to "Off" (e.g. through selecting layers and exporting "stitched" PDF, or through setting initial state in Adobe Reader).

I'm not sure if this is limited to certain companies (tested with Little Lizard King/Styla and Itch to Stitch), and it may be related to #339.

courtneypattison:
Thank you @cfcurtis ! I've created a PR with a possible solution: deep cleaning the hidden layers in the PDF from exporting. It needs a lot more testing and I'm not sure I could make the code robust enough to put in production but it may make sense to go this route like PDFStitcher does.

cfcurtis:
Thanks! I left a comment on that PR, which I think could be a valuable addition. That said, it's actually kind of a different issue - it looks like the code that reads the layers [hard-codes every layer to visible](
https://github.com/Pattern-Projector/pattern-projector/blob/b71c271382882bee4dee111a9b77ade168ddfd14/app/_lib/layers.ts#L23). 

It might be nice to address this on both the reading and the exporting side - right now even if a designer has set a layer's default state to Off (as is common for a calibration grid layer), Pattern Projector will display it as On.

courtneypattison:
Ah, yes! Thank you! Since I was exporting in PP, the visibility was being saved in localStorage so I thought this issue was fixed.

PR #413, opened 2025-12-22
Remove hidden layers from PDF when exporting #412

courtneypattison:
## Pull Request Checklist

## Description

Stitching now removes hidden layers from the PDF when exporting. With the string/stream manipulation in this change, I'm sure many PDFs will have issues, but I wanted to share a possible solution to #412.

## Related Issues

#412 

## TODO

- [ ] Testing on more patterns! I've tested with Wardrobe by Me and Little Lizard King so far.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Line measurement labels match the theme
+- Remove hidden layers from PDF when exporting
 
 ## [1.3.0] - 2025-06-25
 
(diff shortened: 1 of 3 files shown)


cfcurtis:
Wow, that was fast! This is a bit of a rabbit hole, and like you say it might not work for every PDF, so I'm not sure if you'd want it to happen by default. I discovered with PDFstitcher that removing layers isn't always as simple as removing everything between BDC/EMC commands because state changes inside layers persist outside. That means that if there's some change like stroke colour/thickness, font size, etc that happens "in" a layer, it affects the subsequent drawing operations.

My workaround for this was to only remove the "showing" operators in a layer that actually draw stuff and keep the state changes in place, which could probably be replicated here, but it's slow and not ideal.

I'm travelling for holidays right now so I don't have time for a full review/test, but I can do that in January if you'd like.

courtneypattison:
Thank you for the feedback @cfcurtis ! I'll take another pass at this PR with that information in mind and see if I can write something more robust. I'll ping you when I've made those changes to see if you have time to look it over. Thanks again! Happy holidays :)

Issue #416, opened 2026-01-19
Feature proposal: separate control panel


edwardhorsford:
I've been working on a feature (with Claude) idea I had on the weekend - to support a separate control panel that can control the projected view and not require interacting with the ui on the main view.

<img width="50%" alt="Screenshot 2026-01-19 at 09 48 51" src="https://github.com/user-attachments/assets/538b7435-e758-475e-ab8d-9c1f663339ed" />

I've got it working in [this PR on my fork](https://github.com/edwardhorsford/pattern-projector/pull/2) of this repo. Would be very happy if it could get contributed up stream, but as it's mostly LLM written, thought I'd suggest it here first.

## The idea

The main idea is to allow a separate window with key controls that can stay on a normal screen whilst working with the projected screen. I've got it fully working locally - including a new mini-map for navigation. I think this may resolve #393 as it lets you comfortably work from a normal screen whilst keeping pattern projector on the projected view.

## Future ideas

I think it might be possible to get this working such that you could open this control window from a second device - so you could have a phone near the projected screen to control it / move the mini-map. This would be quite easy to do whilst running the app locally, but for a web-deployed thing may need a third party service.

courtneypattison:
@edwardhorsford I love it! The control panel is an amazing use of the BroadcastChannel API. I would definitely consider adding this feature upstream. I just did a quick look over the code and it looks great at first glance. It seems like you're still tinkering with it, but whenever you'd like, please open a PR on the beta branch and I'll start testing it out.

edwardhorsford:
@courtneypattison thank you for the kind words (and for contributing such a great project!).

Happy to open a PR (I should probably test it a bit more first) - though would caution it would probably not have clean commits and it's mostly LLM written. I've not tried deploying yet but will sign up to Vercel so I should be able to get review apps working.

The most complex bit is the mini-map - I got the separate window controls working quickly, but producing a mini map that correctly orientated after flipping and rotating took quite a bit of back and forth. And also challenges around getting the preview image of sufficient contrast (an issue I suspect you've found too). I'm also very glad I did so - I have found the mini-map really useful for navigating and remembering where I am.

Probably for a separate PR but I have an idea around being able to use this mini map to check off pieces as done as you cut them - so you don't forget ones / can remember what you've done. And I reckon it would be super cool if this control panel could be accessed from a phone/tablet.

Issue #393, opened 2025-06-26
Feature request: allow PDF setup on primary display


cfcurtis:
There was a discussion on Facebook about trying to use the following workflow, but being hindered by the calibration warning:

1. Calibrate on projector
2. Move PP window back to primary display
3. Open PDF, select layers, stitch, adjust line weights, etc
4. Move PP window back to projector

FB users talked about how they couldn't do step 3 because of the calibration warning, but of course they don't want to calibrate to their computer screen.

I'm not sure the best way to fix this, and it seems that browser support for detecting a primary display is limited. Perhaps it would make sense to give users a button to hide the calibration warning one time? That way they could move the window, hide the warning, load the PDF etc, and if it pops up again when they move back to the projector then it's legitimate.

hluedeke:
I like this idea. In fact, what I'd probably do is if you dismiss the popup, it shows a small warning icon in the menu somewhere. If you click on the warning icon, it pops the window back up again

courtneypattison:
I can see why people would get frustrated with this flow! When I use extended mode, on my primary display, I press `Project`, open the pattern, do my adjustments, move the window to the extended display, get the calibration warning, click recalibrate, calibrate, and then project.

It's not an obvious or ideal flow for sure, but it works alright. In the future, I'd like the stitching/altering to be separate or the first step in some way. I think that would be nicer, but a lot of work!

I'm hesitant to allow users to dismiss the warning. I could see people thinking they've solved the issue by dismissing the warning. Even if there's still a warning icon, it could be off their table in extended mode.

What about changing the background of the warning's opacity? Similar to how you can see through the main menu now.

hluedeke:
What if it forced you to confirm the close? Like "Are you sure you want to dismiss this?" And explain that it could cause their cuts to be wrong if they dismiss the window. Don't make it easy to close it, but make it possible?

cfcurtis:
Hmm, perhaps this is worth a clarification on the Facebook discussion - they OP said "pattern projector won't let me open the pattern without saving the calibration for the screen on my computer". I just tested on my system (Firefox + Fedora) and I can ignore the warning and open the file - it shouldn't actually block anything, right?

If it's a matter of screen space, could the warning dialog be just on the canvas area so that it's behind the menus? That way it would interfere with projecting, but not setup.

courtneypattison:
[...]

PR #417, opened 2026-01-21
Add separate control panel

edwardhorsford:
## Pull Request Checklist

## Description
This [PR replicates the PR I made on my fork of this repo](https://github.com/edwardhorsford/pattern-projector/pull/2), though with some changes caused by rebasing against the beta branch and removing the settings menu I had added.

The PR adds a secondary "control panel" window. This allows users to control Pattern Projector from a separate screen (like a laptop) while the main window is projected, avoiding the need to interact directly with the projected display.

Screenshot 2026-01-21 at 18 21 06

Replicates most UI functions from the main window to the control panel with the exception of full screen (which had some security issues and didn't work reliably for me). 

Has some two new features including:
* Preview area to let you pan / navigate the pdf
* Controls to pan using buttons.

@courtneypattison Happy to attempt to make any changes you request. Or if you'd like to take this over and make your own changes, that's also cool. I hope it might serve as inspiration for the future even if you decide this PR isn't one you want to merge.

### How it works

**Inter-window Communication** - The control panel and main window communicate via the BroadcastChannel API. A bridge component in the main window listens for actions from the control panel, executes them using the existing transform logic, and broadcasts state updates back. This avoids duplicating any business logic.

**Viewport Calculation** - The preview needs to know what portion of the PDF is visible. The bridge calculates this by transforming screen corners through the inverse of the combined projection matrices, then extracting the bounding box and rotation.

**Thumbnail Generation** - The PDF preview is generated by rendering all stitched pages at 3x the target resolution, applying line thickening based on the line thickness setting, then scaling down with a threshold filter. The result is cached so toggling the preview doesn't trigger re-rendering.

**File Transfer** - Browsers block programmatic file input clicks from other windows, so the control panel reads selected files as ArrayBuffers and transfers them via the broadcast channel. The main window reconstructs the File object and loads it normally.

## Other details

* If border or paper sheet options are enabled, these are shown on the preview
* If the magnify option is enabled, clicking in the preview magnifies the area clicked. Clicking again unmagnifies and reverts the prior window location
* Added options to show or hide the preview as it's a bit slow - unsure how useful this would be in practice though
* Added an option for two sizes of the preview.
* Pan tool lets you pan around. Holding shift pans 10x.
* Line weight settings replicate to the preview - but this isn't working great with higher line weights

## Testing

I've been developing on MacOS with Safari and Arc (Chromium based browser) - and it seems to work. Panning can be a bit slow - but I think the main app was already somewhat slow to pan.

As mentioned in #416 this is mostly LLM coded. I've tested manually but whilst I can code I am not familiar with typescript.

I've only got a handful of pattern pdfs to try and it could do with being tested in more. The preview needs to render from the pdf so it would be good to try other versions. I've tested single and multi-layered, and made sure to regenerate when layers are disabled.

I've not explored the stitch menu.

## Related Issues

#416 and perhaps #393

## TODO

* Changelog - will leave in case you want to make changes / suggest things
* Have not included any UI for opening this control panel - for discussion. In my version, I added a settings menu. For testing it can be accessed by manually opening a new tab and visiting localhost:3000/en/control
* The preview is by far the most complex thing - there could be an argument for pulling it out as a later pr - though for me it's the most useful part of the control panel.
* No translations other than English - I imagine you might want to change some labels, and would only want to translate once settled on them.
* I suspect there are edge cases / things I'm not handling.

## Ideas for the future

* Some of the warning dialogs could be spawned on this window too
* Potential to let a separate phone/tablet load this control panel
* The preview potentially allows features like marking off pieces that have been cut
* I went for a simple single column layout, but could imagine a larger view where there's a larger preview window with smaller controls alongside it


edwardhorsford:
Pushed a fix to some performance issues I noticed - things should be much speedier now.

courtneypattison:
Thank you so much for this, @edwardhorsford! It’s an impressive amount of work.

After thinking about the control panel idea for a few days, I’m actually leaning toward a simpler screen mirroring option. Instead of maintaining a completely different UI for the control panel (which adds a lot of complexity and duplication), I think it’d be much more maintainable to just mirror the existing UI across tabs.

Since we already have the "zoom out" view—which is essentially a minimap—we can focus our energy on making the main UI better for everyone. For example, if we move toward a responsive, reflowing, or draggable menu, that solves the problem for both the "controller" tab and for people just using the app on a single screen or mobile device. (Though I know I still need to fix how zoom out works currently in #379!)

A few thoughts on your ideas:

Separate Devices: It would be awesome to get this working on a phone/tablet! I’m hesitant to host anything that touches user data, as I want to keep PP strictly client-side. But even just an educational resource or a local-network sync guide would be a huge win.

Marking off pieces: A lot of people use the line tool for this. I actually do the opposite: I draw lines on the grainlines of pieces I need to cut, then delete them as I go. It helps me align grainlines with my cutting mat and keeps a "countdown" of what’s left.

Mobile/Layout: I see why you went for the single-column layout—it’s much better for mobile. I’d love to take that inspiration and apply it to a responsive design for the main site, rather than maintaining two separate versions.

What do you think about focusing the "bridge" logic on syncing the state between two identical tabs instead? That way, we get the functionality you're looking for with a fraction of the code to maintain.

edwardhorsford:
Hey @courtneypattison, sorry for the delay - I have been distracted by projects.

I totally get your reservation about the code duplication and can see the simplicity of mirroring the view. Though having used my fork for real on a project, I'm really pleased with the utility of the mini map - I don't think I'd want to go back to just the regular view. I'd actually quite like *both* the regular map and mini map at the same time in my control window.

To throw a potential curveball, I have half a reckon that one window should be for controls, and the second window for *just* the projected view - minus most of the current UI. So not two mirrored views, but slimming down the projected thing to just being display only. Or perhaps a system where you accept there are two windows and you make it configurable what is on each.

---

Marking off pieces - [I did this here](https://github.com/edwardhorsford/pattern-projector/pull/4) and really liked using it. Pressing 'x' triggered a tick in the middle of the screen.

>What do you think about focusing the "bridge" logic on syncing the state between two identical tabs instead? That way, we get the functionality you're looking for with a fraction of the code to maintain.

I'm noodling on some more window ideas and features, but I'm not sure I'd go the direction of having two identical windows - so I'll probably just do that on my fork. Though I might play around with trying to reduce code duplication to enable a given ui control to exist in different contexts/windows. I'll close this PR as it doesn't sound suitable for this project - hopefully it might serve as some inspiration for the future though.

PR #420, opened 2026-03-06
Loupe line ends

edwardhorsford:
Adds a loupe view shown when positioning line ends. Could probably do with some tweaking to make small adjustments.

loupe-line-ends


edwardhorsford:
Sorry, accidental PR

PR #421, opened 2026-03-27
Replace time-acceleration with modifier-key granularity for arrow mov…

IoannisNezis:
Hey @courtneypattison,

my mom discrovered your amazing software and is a huge fan.
It saves her so much time and pain! Its extremely usefull.

I was delighted to see that this is a opensource project that's freely available.
I'm a software engineer myself and want to contribute to your project.

There a few things i noticed and would like to improve.
Starting with this one:

## Pull Request Checklist

### Arrow key movement: modifier-based granularity + diagonal support

https://github.com/user-attachments/assets/750efc40-9594-43ad-91a1-1e755fe9ffd1

#### What changed

Replaces the time-acceleration approach (cycling through a pixel list the longer an arrow key is held) with explicit granularity levels controlled by modifier keys:

| Keys                           | Step size |
|-----------------------------|---------------|
| Arrow                        | 1 px          |
| Shift + Arrow            | 10 px       |
| Alt + Arrow               | 0.5 px      |
| Ctrl + Arrow             | 0.1 px       |
| Shift + Ctrl + Arrow | 100 px     |
| Shift + Alt + Arrow  | 0.05 px    |

#### Behaviour improvements

- **Seamless granularity switching** - adding or removing a modifier key while holding an arrow key transitions immediately, without releasing and re-pressing the arrow key.
- **Diagonal movement** - holding two arrow keys simultaneously moves on both axes at once.
- **No motion interruption on quick key switches** - held keys are tracked as a set; the repeat interval is shared across all held arrows and only stops when all are released.

#### Implementation notes

Movement is driven by a self-managed `setInterval` (30 ms, after a 400 ms initial delay) rather than relying on browser `keydown` auto-repeat. This fixes the browser halting repeated `keydown` events when a modifier key is added while an arrow key is held. Modifier and arrow state are stored in refs so the interval callback always reads current values without restarting. Held-key state is cleared on deactivation to prevent ghost movement if the hook is toggled while keys are physically held.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Enabled measuring when magnified for precision measuring
 - Pressing an arrow button once will move the pattern by 1/8" or 2mm
 - Slow down speed of pattern movement with arrow keys
+- Arrow key movement uses modifier keys for granularity (Shift = coarse, Ctrl = fine, Alt = finer, combinations for extremes); supports diagonal movement and seamless granularity switching without releasing the arrow key
 
 ## [1.3.0] - 2025-06-25
 
--- a/app/_hooks/use-prog-arrow-key-handler.ts
+++ b/app/_hooks/use-prog-arrow-key-handler.ts
@@ -1,61 +1,145 @@
 import { KeyCode } from "@/_lib/key-code";
-import { useCallback, useEffect, useState } from "react";
+import { useCallback, useEffect, useRef } from "react";
 
-const PIXEL_LIST = [1, 10, 20, 40];
+export interface ArrowKeyGranularity {
+  normal: number;
+  shift: number;
+  ctrl: number;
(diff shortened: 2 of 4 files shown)


courtneypattison:
@IoannisNezis Apologies for taking so long to respond.

I do like the idea of being able to change the granularity of movement. Even for moving the whole pattern when projecting. My only concern it that this PR would make the keyboard versus touchscreen experience diverge quite a bit. Right now the move tool is the main way for touchscreen users to fine tune the corners on the calibration grid and that uses the time-acceleration method. Do you have any ideas for how to make both experiences similar? Maybe have an option in the move tool to adjust granularity? I'm not sure how that would look though.

Thank you so much for your contribution!

edwardhorsford:
In my fork of the project I've generally adopted fine movements as standard, with shift being 10x - which is what Adobe Illustrator and many other apps tend to do. I'm not sure it needs 6 levels of adjustment - 0.5mm and 5mm has been plenty for me.

PR #425, opened 2026-04-13
Upstream line quality improvements

edwardhorsford:
## Summary

This upstreams some of the work I did in my fork to improve the line quality.

---

### Commit 1 — Push darks and contrast boost

Introduces a `push-darks` SVG filter (gamma=2 feComponentTransfer) and a `contrast(1.5)` step applied to every PDF canvas render. These are always applied, even at zero line thickness, because they improve the raw PDF output: grey anti-aliased edges are pushed back toward black, producing sharper, cleaner lines.

On Safari, where SVG filter references cannot be applied to canvas contexts, an equivalent pixel-level LUT (`enhanceLineQualityFast`) is used instead.

---

### Commit 2 — feColorMatrix recolour

Replaces the Green theme's CSS filter chain (`invert sepia saturate hue-rotate brightness`) with a single `#recolor` feColorMatrix SVG filter that directly maps luminance to the target colour:

```
output = targetColour × (1 − luminance(input))
```

Black input → full target colour. White input → black. Grey inputs → proportionally dimmer shades. This gives exact, consistent colour across browsers, and makes the target colour configurable as a hex value rather than an approximation via hue-rotate. I've picked a light greenish hue similar to the brightest in the existing Pattern Projector, but you might want to tweak it.

On Chrome/Firefox, the recolour is applied via `ctx.filter` at canvas draw time. On Safari, pixel-level processing (`recolourImageData`) handles it.

### DIFF
--- a/app/[locale]/calibrate/page.tsx
+++ b/app/[locale]/calibrate/page.tsx
@@ -27,6 +27,8 @@ import {
   DisplaySettings,
   getDefaultDisplaySettings,
   isDarkTheme,
+  isColourTheme,
+  strokeColor,
   themeFilter,
   Theme,
 } from "@/_lib/display-settings";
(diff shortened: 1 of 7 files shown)


courtneypattison:
Awesome, thank you @edwardhorsford !

edwardhorsford:
@courtneypattison no worries. Hopefully it makes sense. I had to tinker quite a bit - esp trying to reduce flashes of un-inverted styles.

I held off moving to workers in this, though Claude suggested it several times.

